PR #1, opened 2023-08-15
Huge Japanese translation update

logue:
[FIRE-33076](https://jira.firestormviewer.org/browse/FIRE-33076) Japanese Translation Update


Ansariel:
This change unfortunately has the same issues as the initial one contributed to the LL viewer: Please do not unnecessarily change the formatting and indendation of the files as it will result in lots and lots of merge conflicts with the LL viewer.

Some files are not necessarily required, e.g. half of the files for the AnsaStorm skin are not required as the translations pulled from the default skin work fine. What is actually needed for AnsaStorm can be determined by checking the "de" folder.

logue:
> This change unfortunately has the same issues as the initial one contributed to the LL viewer: Please do not unnecessarily change the formatting and indendation of the files as it will result in lots and lots of merge conflicts with the LL viewer.

I don't know if it's SDL or expat specifications, but if the character contains a tab, it will be garbled into "□". I had this problem in my own project using the same set of techs, and it was happening in Firestorm as well.

I made it based on the original `en` directory, but some files had a mixture of tab indentation and space indentation, so I unified the indentation with two spaces. 

Which indentation format should I use here?

> Some files are not necessarily required, e.g. half of the files for the AnsaStorm skin are not required as the translations pulled from the default skin work fine. What is actually needed for AnsaStorm can be determined by checking the "de" folder.

The file structure was different depending on the language directory, so I translated all the files just in case, but is it correct to say that I should refer to the directory in 'de'?

Ansariel:
> Which indentation format should I use here?

Stick to the indentation and formatting prevelant in the files. Having a completely different file is a nightmare when merging upstream changes.

> The file structure was different depending on the language directory, so I translated all the files just in case, but is it correct to say that I should refer to the directory in 'de'?

For AnsaStorm yes. Pretty sure "pl" is far more complete in terms of all skins.

logue:
> Stick to the indentation and formatting prevelant in the files.
So what is the proper indentation setting?

At least, `floater_associate_listing.xml` sees indentation as 3 spaces, whereas `rlva_strings.xml` sees it as 2 tabs and `panel_region_experiences.xml` sees it as 2 spaces.

floater_associate_listing.xml
image

rlva_strings.xml
image

panel_region_experiences.xml
image

checked the [Firestorm Wiki](https://wiki.firestormviewer.org/) before starting work, but I couldn't find any documentation about indentation conventions.

I used XML Tools because I needed some format conventions to work with.

> For AnsaStorm yes. Pretty sure "pl" is far more complete in terms of all skins.
OK, Now delete unnecessary files.

Ansariel:
The format convention is: Don't mess with the formatting coming from upstream because that will create a huge mess when merging and makes me simply discard either version.

logue:
understood.
As far as I read upstream, it seems to be tab indentation, so I will standardize it.

However, as I mentioned earlier, the following case may cause garbled characters, so it is set to one line:

```xml
	<text name="desc txt">
		説明：
	</text>
```

becomes

```xml
	<text name="desc txt">説明：</text>
```

logue:
@Ansariel

I fixed the indentation.

Ansariel:
Closing this PR since we have merged a big update from LL in the meantime and this PR now has dozens of conflicts. Additionally, there is another PR on LL's side waiting to get merged as well, which will just create even more conflicts.

logue:
@Ansariel OK, 
Please let me know by DM when you are ready to accept. (Updates have slowed down, etc.)
Either way, it's definitely going to be a major overhaul.
Then work again.

PR #3, opened 2023-08-26
Aleric 00 which lod

AlericInglewood:
The saga continuous.

### DIFF
--- a/indra/newview/llmodelpreview.cpp
+++ b/indra/newview/llmodelpreview.cpp
@@ -2523,7 +2523,7 @@ void LLModelPreview::genMeshOptimizerLODs(S32 which_lod, S32 meshopt_mode, U32 d
         out << "Invalid level of detail: " << which_lod;
         LL_WARNS() << out.str() << LL_ENDL;
         LLFloaterModelPreview::addStringToLog(out, true); // <FS:Beq/> if you don't flash the log tab on error when do you?
-        assert(which_lod >= -1 && which_lod < LLModel::NUM_LODS); // <FS:Ale/> use the correct variable (which_lod).
+        assert(which_lod >= -1 && which_lod < LLModel::NUM_LODS); // <FS:PR-Aleric/> use the correct variable (which_lod).
         return;
     }
 
--- a/indra/newview/llmodelpreview.cpp
+++ b/indra/newview/llmodelpreview.cpp
@@ -2523,7 +2523,7 @@ void LLModelPreview::genMeshOptimizerLODs(S32 which_lod, S32 meshopt_mode, U32 d
         out << "Invalid level of detail: " << which_lod;
         LL_WARNS() << out.str() << LL_ENDL;
         LLFloaterModelPreview::addStringToLog(out, true); // <FS:Beq/> if you don't flash the log tab on error when do you?
-        assert(lod >= -1 && lod < LLModel::NUM_LODS);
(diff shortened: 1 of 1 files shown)


indra/newview/llmodelpreview.cpp
```diff
@@ -2523,7 +2523,7 @@ void LLModelPreview::genMeshOptimizerLODs(S32 which_lod, S32 meshopt_mode, U32 d
         out << "Invalid level of detail: " << which_lod;
         LL_WARNS() << out.str() << LL_ENDL;
         LLFloaterModelPreview::addStringToLog(out, true); // <FS:Beq/> if you don't flash the log tab on error when do you?
```
AlericInglewood, line 2525: This is an added (review) comment.

beqjanus: With this, I'm wondering why the assert is even there; it serves no purpose and adds nothing but technical debt (as demonstrated by this bug). The better solution is, I think, to remove the line, Which in FS coding terms means commenting it out with the tags. I'll get back to you on the preferred tag ASAP.

beqjanus: if we do want an assert there then assert(false) as you mentioned elsewhere would be less fragile.

AlericInglewood: Personally (in my own projects anyway) I prefer an assert (or whatever hard abort with stack trace) whenever an undesirable situation occurs. Where 'undesirable' can range from "unexpected" to "not understood" to "should never happen" to "can't be recovered from". My reasoning is that if anything happens that wasn't anticipated then we don't know what will happen next and if that is a crash or just undesirable behavior then we need to know when the state of the program was inconsistent as early as possible. I am aware of the reasoning of certain (open source) developers that in the same situation (where I'd add an assert) they want to attempt to recover and not crash the program, because that would reflect bad on the reputation of the project (e.g. people rather use a viewer with graphical artifacts that one that crashes precisely where the bug is; and might switch to another viewer).

Given that Release and even RelWithDebInfo doesn't compile asserts in the latter is already achieved. However, I am compiling and USING the viewer at all times with assert's compiled in, exactly for this reason: gimmi a crash - any - just one time - and I'll fix it.

In my community (in opensim) *everyone* is currently crashing during teleports about 1 every 4 times; I have crashed myself too many times, but haven't teleported a lot since I started to use the self-compiled viewer (so no stack trace of that yet). But I guess that's besides the point here ;)

TL;DR Please keep the assert - though assert(false) would be totally fine.

indra/newview/llmodelpreview.cpp
```diff
@@ -2523,7 +2523,7 @@ void LLModelPreview::genMeshOptimizerLODs(S32 which_lod, S32 meshopt_mode, U32 d
...
-        assert(lod >= -1 && lod < LLModel::NUM_LODS);
+        assert(which_lod >= -1 && which_lod < LLModel::NUM_LODS); // <FS:Ale/> use the correct variable (which_lod).
```
AlericInglewood: Just want to see what happens if I click 'Start a review'.

AlericInglewood: Looks like I am the one doing the "review" after that, and I have to Finish the review myself.
Look good! Though I'd use <FS:PR-Aleric/>, what you think? ;)

PR #4, opened 2023-08-27
Aleric 01 f pic

AlericInglewood:
I didn't like that `fs-build-variables` dragged in a `-fPIC` as CXXFLAGS because,
1) that is not how one should do that when using cmake (see this patch for how to do that).
2) that causes ALL files to be compiled with that flag, also normal viewer files that don't need it.
3) it isn't necessary for amd64.
Hence, I removed that flag locally.

Turns out this (llrender) is the *only* thing that *does* need to be compiled with -fPIC (despite being on amd64) because...

So, added this line to make it so.

Needless to say that if you keep -fPIC in `fs-build-variables` anyway then this change has no effect,
but I'd appreciate it if it is accepted since it makes sense ;).

### DIFF
--- a/indra/llrender/CMakeLists.txt
+++ b/indra/llrender/CMakeLists.txt
@@ -88,6 +88,8 @@ endif (BUILD_HEADLESS)
 add_library (llrender ${llrender_SOURCE_FILES})
 target_include_directories( llrender  INTERFACE   ${CMAKE_CURRENT_SOURCE_DIR})
 
+set_property(TARGET llrender PROPERTY POSITION_INDEPENDENT_CODE ON)
+
 # Libraries on which this library depends, needed for Linux builds
 # Sort by high-level to low-level
 target_link_libraries(llrender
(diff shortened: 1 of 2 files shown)


AlericInglewood:
Argh.  Something went wrong - my intention was to only do a pull request for the one-liner in indra/llrender/CMakeLists.txt but somehow a commit from someone else got dragged in :/.

AlericInglewood:
A merge commit got in there... lemme try again.

yaynstuff:
There are cases where you might want to use fPIC, please explain why its not needed on amd64. Some reasons why one might want to use it in a project (note: not all may apply to the viewer, but included for thoroughness):

- Building Shared Libraries: Shared libraries (also known as dynamic link libraries) are compiled code that can be loaded into memory at runtime by multiple programs. Since these libraries can be loaded at different memory addresses for different processes, the code needs to be position-independent. This is especially important on 64-bit systems where the memory layout is more complex.

- Address Space Layout Randomization (ASLR): Modern operating systems, including Linux, use ASLR to enhance security by randomly arranging the memory locations of various program components. Position-independent code is more compatible with ASLR as it can be loaded at different memory addresses without causing relocation issues.

- Dynamic Loading: In some cases, you might be using mechanisms like dlopen in C or C++ to dynamically load shared libraries at runtime. Position-independent code is essential for ensuring that the code can be loaded and executed properly regardless of where it's placed in memory.

- Plugins: If you are developing a system that supports dynamic loading of plugins, such as a text editor that allows third-party extensions, you would want these plugins to be compiled with -fPIC to ensure they can be loaded and executed correctly in the host process's memory space.

-  Avoiding Text Relocations: Position-independent code reduces the need for text relocations, which are costly operations that involve modifying the code section of a shared library at load time. Minimizing text relocations improves performance.

So you have taken all the scenarios into account and are quite certain its a good idea? Right off the bat, one thing that comes to mind is that the viewer does open some shared libaries at runtime. The viewer has been fine with fPIC forever now, just wondering what the driving idea is behind wanting to remove it.

PR #5, opened 2023-08-28
llrender must be compiled with -fPIC.

AlericInglewood:
I didn't like that fs-build-variables dragged in a -fPIC as CXXFLAGS because,

1) that is not how one should do that when using cmake (see this patch for how to do that).
2) that causes ALL files to be compiled with that flag, also normal viewer files that don't need it.
3) it isn't necessary for amd64.

Hence, I removed that flag locally.

Turns out this (one file of llrender) is the only thing that *does* need to be compiled with -fPIC (despite being on amd64) because [...].

So, added this line to make it so.

Needless to say that if you keep -fPIC in fs-build-variables anyway then this change has no effect,
but I'd appreciate it if it is accepted since it makes sense ;).

### DIFF
--- a/indra/llrender/CMakeLists.txt
+++ b/indra/llrender/CMakeLists.txt
@@ -88,6 +88,8 @@ endif (BUILD_HEADLESS)
 add_library (llrender ${llrender_SOURCE_FILES})
 target_include_directories( llrender  INTERFACE   ${CMAKE_CURRENT_SOURCE_DIR})
 
+set_property(TARGET llrender PROPERTY POSITION_INDEPENDENT_CODE ON)
+
 # Libraries on which this library depends, needed for Linux builds
 # Sort by high-level to low-level
 target_link_libraries(llrender


yaynstuff:
There are cases where you might want to use fPIC, please explain why its not needed on amd64. Some reasons why one might want to use it in a project (note: not all may apply to the viewer, but included for thoroughness):

    Building Shared Libraries: Shared libraries (also known as dynamic link libraries) are compiled code that can be loaded into memory at runtime by multiple programs. Since these libraries can be loaded at different memory addresses for different processes, the code needs to be position-independent. This is especially important on 64-bit systems where the memory layout is more complex.

    Address Space Layout Randomization (ASLR): Modern operating systems, including Linux, use ASLR to enhance security by randomly arranging the memory locations of various program components. Position-independent code is more compatible with ASLR as it can be loaded at different memory addresses without causing relocation issues.

    Dynamic Loading: In some cases, you might be using mechanisms like dlopen in C or C++ to dynamically load shared libraries at runtime. Position-independent code is essential for ensuring that the code can be loaded and executed properly regardless of where it's placed in memory.

    Plugins: If you are developing a system that supports dynamic loading of plugins, such as a text editor that allows third-party extensions, you would want these plugins to be compiled with -fPIC to ensure they can be loaded and executed correctly in the host process's memory space.

    Avoiding Text Relocations: Position-independent code reduces the need for text relocations, which are costly operations that involve modifying the code section of a shared library at load time. Minimizing text relocations improves performance.

So you have taken all the scenarios into account and are quite certain its a good idea? Right off the bat, one thing that comes to mind is that the viewer does open some shared libaries at runtime.  See also "Avoiding Text Relocations" and some of the other points.

yaynstuff:
Regarding "not needed on x64" there's no real golden rule on that or any overwhelming reason not to use it that I can see.

While it's true that some aspects of 64-bit systems can mitigate the need for position-independent code (-fPIC), there are still cases where using -fPIC remains important even on 64-bit systems. Here are a few points to consider:

- Address Space Layout Randomization (ASLR): ASLR is used on both 32-bit and 64-bit systems to randomize the memory addresses of various components. While 64-bit systems do have larger address spaces, they still benefit from ASLR to enhance security. Position-independent code is helpful here to ensure that code can be loaded at various randomized memory locations.

- Shared Libraries: Shared libraries are used extensively in both 32-bit and 64-bit systems. If you are building shared libraries for use across different processes or applications, you'll still need -fPIC to ensure that the code can be loaded properly into different memory spaces.

- Compatibility: Even though 64-bit systems have larger address spaces, not all code or systems are exclusively 64-bit. You might need to ensure that your code works on both 32-bit and 64-bit systems, and using -fPIC can provide this compatibility.

- Cross-Platform Development: If you're developing code that's intended to run on multiple platforms, using -fPIC can help ensure consistent behavior and compatibility across those platforms, regardless of their architecture.

- Binary Interoperability: If your code interacts with other components or libraries that are compiled with position-independent code, it's generally better to maintain consistency by using -fPIC in your own code as well.

While some of the technical reasons for using -fPIC might be less critical on 64-bit systems, it's still considered a good practice to use it when building shared libraries or code that needs to be dynamically loaded. It's always a balance between optimizing for specific scenarios and ensuring compatibility and best practices across different environments.

beqjanus:
I'm not totally convinced by any of the arguments one way or the other :-) . Historically I have avoided PIC in performance code because of the interaction with pipelining and speculative execution. However, given that all the 3p libs are built with PIC it seems marginal. All that said, I am merging the change because it allows you to make the choice locally.

PR #9, opened 2023-08-28
Add CMAKE_CONFIG `Debug` back.

AlericInglewood:


Ansariel:
Debug builds are definitely NOT coming back.

AlericInglewood:
> Debug builds are definitely NOT coming back.

Hi Ansariel :)
I compile without optimization (and more importantly, without `-DNDEBUG`) all the time, and use the viewer like that.
Of course I can maintain this locally, but I'm curious: why is it considered bad to support using `Debug` besides `RelWithDebInfo`?

Here is how I currently configure the viewer:
```
autobuild configure -A 64 -c DebugFS_open -- --fmodstudio --compiler-cache  -DCMAKE_CXX_COMPILER=/usr/bin/g++ -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON
```
with a few environment variables as well. Works like a charm :/

Ansariel:
I wouldn't call 1 FPS "like a charm". There is no reason to run a debug build over a release build with debug symbols.

PR #11, opened 2023-08-28
Fix "duplicated copyright" error.

AlericInglewood:
This solves a 'Duplicate copyrights for zlib-ng' problem. The reason turned out to be a mysterious restart of reading the output pipe of the command

  copyrights=autobuild('install', '--copyrights', '--install-dir', args.install_dir).

The output of that command is:
```
  Copyright (c) 2023, The Phoenix Firestorm Project, Inc.
  SDL2: Copyright (C) 1997-2022 Sam Lantinga
  ...
  zlib-ng: Copyright (C) 1995-2013 Jean-loup Gailly and Mark Adler
```
where the first line is consumed by `viewer_copyright = copyrights.readline()`.

But because the stream is read twice, it looks as if we have:
```
  ...
  zlib-ng: Copyright (C) 1995-2013 Jean-loup Gailly and Mark Adler
  Copyright (c) 2023, The Phoenix Firestorm Project, Inc.
  SDL2: Copyright (C) 1997-2022 Sam Lantinga
  ...
  zlib-ng: Copyright (C) 1995-2013 Jean-loup Gailly and Mark Adler
```
and now that first `Copyright (c) 2023, The Phoenix Firestorm Project, Inc.` line is added to the copyright line(s) of zlib-ng (because it doesn't begin with `packagename:`), which then results in "two" zlib-ng packages with a "different" copyright line.

I am not sure if the code was buggy or that python3 on Arch is buggy, but this code is shorter and easier to understand, too.

### DIFF
--- a/scripts/packages-formatter.py
+++ b/scripts/packages-formatter.py
@@ -86,36 +86,29 @@ viewer_copyright = copyrights.readline() # first line is the copyright for the v
 # Two different autobuild outputs, but we treat them essentially the same way:
 # populating each into a dict; each a subdict of 'info'.
 for key, rawdata in ("versions", versions), ("copyrights", copyrights):
-    lines = iter(rawdata)
-    try:
-        line = next(lines)
-    except StopIteration:
-        # rawdata is completely empty? okay...
-        pass
-    else:
+# <FS:Al> The trick with re-entering `for line in lines:` doesn't work: it causes all lines to be read twice (on my machine).
+    # Just read the output once and use simple(r) logic.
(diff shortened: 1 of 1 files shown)


AlericInglewood:
I retract this one... it still doesn't work.
The lines are STILL read twice and the only reason that I'm no longer getting a duplicated copyright is because this way multi-line copyright lines don't work anymore...

Ansariel:
This is not an error in this script, but in some 3p package. I had that happen some months ago.

PR #12, opened 2023-08-29
Workaround for repeated viewer copyright.

AlericInglewood:
This happens when autobuild outputs everything twice. I can't figure out why it does that, but it does.

The problem of that is that the second 'viewer copyright' line is then appended to the copyright line(s) of the last packages, causing the error "Duplicated copyright" (thinking we have two packages with the same name but a different copyright).

### DIFF
--- a/scripts/packages-formatter.py
+++ b/scripts/packages-formatter.py
@@ -109,7 +109,7 @@ for key, rawdata in ("versions", versions), ("copyrights", copyrights):
                     # we hit the start of the next package data
                     add_info(key, pkg, pkg_lines)
                     break
-                else:
+                elif line != viewer_copyright:  # <FS:PR-Aleric> skip possibly repeated viewer copyright line(s).
                     # no package prefix: additional line for same package
                     pkg_lines.append(line.rstrip())
             else:


AlericInglewood:
Thus, in short:
```
$ autobuild install --copyright --install-dir /opt/secondlife/viewers/firestorm/phoenix-firestorm-git/build-linux-x86_64/packages | grep -E '(SDL2|zlib-ng|Firestorm)'
Copyright (c) 2024, The Phoenix Firestorm Project, Inc.
SDL2: Copyright (C) 1997-2022 Sam Lantinga
zlib-ng: Copyright (C) 1995-2013 Jean-loup Gailly and Mark Adler
Copyright (c) 2024, The Phoenix Firestorm Project, Inc.                     <--- this line gets added to the previous one
SDL2: Copyright (C) 1997-2022 Sam Lantinga
zlib-ng: Copyright (C) 1995-2013 Jean-loup Gailly and Mark Adler
```

Ansariel:
As mention in the other PR: This is not a problem in this script, but in a 3p package.

AlericInglewood:
> As mention in the other PR: This is not a problem in this script, but in a 3p package.

No it is not a problem in any 3p package. See my explanation above and the output of autobuild.
autobuild repeats all output twice (maybe because it thinks all packages are installed twice, maybe because it is), but its definitely ALL packages and a 100% repeat of the expected output.

The PROBLEM then is that the "viewer copyright" line is consumed the first time, but (obviously) not the second time. I addressed this in this one liner; it has nothing to do with a 3p package, even though you ran into the same error before in which case it *was* a problem with a 3p package.

Ansariel:
Of course it is a problem in one of the 3p packages because it happened to me a while ago. It's not a problem in this script. So I'll just close this PR.

AlericInglewood:
Which package is the problem then?  And how exactly does it cause all the output of autobuild to be duplicated?
I mean, you ran into this before, so it would be nice to explain it to me.

PR #14, opened 2024-01-11
Add linux use systemlibs

AiraYumi:
There are some packages that may be able to use packages installed on your system.
Change it to use the packages installed on your system.

This only affects Linux and does not affect Windows or Mac environments.

This PR is currently a draft as we would like to hear opinions from developers.
The number of eligible packages may increase or decrease.

### DIFF
--- a/indra/cmake/00-Common.cmake
+++ b/indra/cmake/00-Common.cmake
@@ -135,6 +135,10 @@ endif (WINDOWS)
 if (LINUX)
   set(CMAKE_SKIP_RPATH TRUE)
 
+  # Configure whether to use the library or prebuilt library once installed on the system.
+  # fltk
+  set(FLTK_USE_SYSTEMLIBS FALSE)
+
   # <FS:ND/>
   # And another hack for FORTIFY_SOURCE. Some distributions (for example Gentoo) define FORTIFY_SOURCE by default.
   # Check if this is the case, if yes, do not define it again.
--- a/indra/cmake/APR.cmake
+++ b/indra/cmake/APR.cmake
@@ -5,8 +5,14 @@ include_guard()
 
 add_library( ll::apr INTERFACE IMPORTED )
 
-use_system_binary( apr apr-util )
-use_prebuilt_binary(apr_suite)
+if (NOT LINUX)
+  use_system_binary( apr apr-util )
+  use_prebuilt_binary(apr_suite)
+else ()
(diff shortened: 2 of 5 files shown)


yaynstuff:
Thank you for taking the time to share your work with us and submitting it for consideration.

The reason why we use the bundled version of APR rather than a system wide one is because LL have forked their own instance of APR and have heavily modified it - LL have introduced custom functionality into that library which the viewer depends on, and which does not exist in stock standard versions of the APR library.

Have a look at **indra/llcommon/llprocess.cpp**

Around line 566:
```
#if ! defined(APR_HAS_PROCATTR_CONSTRAIN_HANDLE_SET)
        // Our special preprocessor symbol isn't even defined -- wrong APR
        LL_WARNS("LLProcess") << "This version of APR lacks Linden "
                                                  << "apr_procattr_constrain_handle_set() extension" << LL_ENDL;
#else
```

`llcommon/llprocess.cpp:646:             LL_WARNS("LLProcess") << "This version of APR lacks Linden apr_procattr_autokill_set() extension" << LL_ENDL;`

You can find LL's fork of APR here:  https://github.com/secondlife/3p-apr_suite 

For FLTK, this library not always present on peoples systems and making the viewer depend on the system wide version being present could cause breakage for a lot of folks - while its easily solved by telling those affected to install the package, it would result in more load on our support team and no doubt people creating Jira tickets complaining that the viewer wont load.

yaynstuff:
Also there are some libraries that are just not safe to use the system wide version. OpenSSL is also a good example, OpenSSL are known for making API changes which often require viewer source code changes to make it compatible.

beqjanus:
Hi @AiraYumi 
While I am in favour of changes that allow self-compilers and package maintainers to have a slightly easier job, if it does that at the cost of our own build complexity, then we can't accept the PR. As noted by @yaynstuff, there are various reasons why we are tied to the LL-specific builds, and I am wary of blurring that relationship if it leads to possible issues.
If the changes are structured so that the prebuilt non-system libraries are used by default, with a build time switch that allows the system libs to be used where appropriate, then the changes would be more appropriate and we would be better able to consider them.

AiraYumi:
Thank you for your opinion.
I did a little research after that, and it seems that Ubuntu also installs static libraries. I wasn't able to determine how to statically link with cmake, but it might be able to resolve some concerns. I would also like to provide a way to choose whether to use system libraries or not.

PR #15, opened 2024-02-01
Update grids.xml

nbtafelberg:
Add Wolf territories grid which is the largest grid by land area (even bigger than SL)

### DIFF
--- a/indra/newview/app_settings/grids.xml
+++ b/indra/newview/app_settings/grids.xml
@@ -1,5 +1,32 @@
 <llsd>
     <map>
+            <key>grid.wolfterritories.org:8002</key>
+        <map>
+        <key>LastModified</key>
+            <date>2024-01-01T11:41:00.95Z</date>
+        <key>gridname</key>
+            <string>Wolf Territories Grid</string>
+        <key>gridnick</key>
+            <string>WolfGrid</string>
+        <key>login_identifier_types</key>
+            <array>
+                <string>agent</string>
+                <string>account</string>
+            </array>
+        <key>loginpage</key>
+            <string>https://www.wolf-grid.com</string>
+        <key>loginuri</key>
+            <array>
+                <string>http://grid.wolfterritories.org:8002</string>
+            </array>
+        <key>name</key>
+            <string>grid.wolfterritories.org:8002</string>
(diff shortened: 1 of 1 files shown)


beqjanus:
Not sure what to do with this. We stated about 2 years ago, maybe more now, that we were no longer maintaining grids.xml. It is not only a very poor way to discover new grids but it adds a burden of maintenance upon us that we cannot support. The correct way to manage a grid "entry" in FS now is to provide your users with the viewer URL link that adds the grid automatically. That said the grid list is woefully out of date and I have not deleted it so I should either remove it completely, or do a manual one time refresh of the top N grids. What I would really like is to have OpenSim (where I use the term in the loosest sense meaning the collective of users of opensim in all its forms) has a de-facto grid directory that the viewer (any viewer) could then build a suitable grid list from.

aiaustin:
Would it be helpful if I did a little research on this just to get a tidy grids.xml file that can be used as a clean base for Firestorm 7.*. We could also clear this PR if its accepted that Wolf Territories has the top traffic figures at present according to Hypergrid Business's OpenSim statistics.   A "non-controversial" approach could be taken of just removing explicitly the deprecated and closed down sites and where even DNS addresses are no longer working (only three of them I find).

Initial research at https://blog.inf.ed.ac.uk/atate/2024/05/13/firestorm-grids-xml/

aiaustin:
I add my grids.xml here in case its useful (had to add a .txt on the end to get it to uload here). But I totally understand Beq's position.  I guess hyperica.com was this sort of thing - a hypergrid locations directory originally curated by Maria Kolorow and now Fred Beckhusen. But its not easy to navigate or use.

[grids.xml.txt](https://github.com/FirestormViewer/phoenix-firestorm/files/15353999/grids.xml.txt)

beqjanus:
Hi @nbtafelberg I'm closing this as I am close to completing a revamp of the grid system. Here's a preview of what it should look like (filtered using "ter"). The grids.xml itself is going to be stripped back to nearly nothing. Users can then add grids simply using that popup which is auto-generated from Maria Korolov's data and should refresh once a month as her reports do.
7a7ffdafddd3f5583f5cae5fe3f5e80e

PR #23, opened 2024-05-05
fix RelWithDebInfo build error: 'comboTexGen': undeclared identifier

tapple:
This fixes the only build error I encountered trying to run `autobuild build -A 64 -c RelWithDebInfoFS_open`:
```
E:\cabbage\sl\phoenix-firestorm\indra\newview\fspanelface.cpp(1152,3): error C2065: 'comboTexGen': undeclared identifie r [E:\cabbage\sl\phoenix-firestorm\build-vc170-64\newview\firestorm-bin.vcxproj]
```
all the `getChild()` calls here are null-checked with an `if`, so I'm not sure why there's any llassert here

### DIFF
--- a/indra/newview/fspanelface.cpp
+++ b/indra/newview/fspanelface.cpp
@@ -1149,7 +1149,6 @@ struct FSPanelFaceSetTEFunctor : public LLSelectedTEFunctor
 
 		bool align_planar = mPanel->mCheckPlanarAlign->get();
 
-		llassert(comboTexGen);
 		llassert(object);
 
 		if (ctrlTexScaleS)


tapple:
nevermind. there were way more build errors after this one

tapple:
Although the viewer builds after this change, it still crashed: 
```
2024-05-05T19:57:33Z ERROR # llui/llpanel.cpp(413) LLPanel::fromXML : ASSERT (panelp)
```

beqjanus:
What platform do you compile on @tapple . If you are building on Windows then RelWithDebInfo is not that useful as you get the same result running the unstripped binary firestorm-bin.exe from the build folder. Not sure on other platforms though.

Can you see which panel is null on startup? llpanel is a base class so the problem is presumably a derived class and you should be able to see that in the stack trace.

PR #26, opened 2024-05-18
One off update to app_settings grids.xml

aiaustin:
Update as described at https://blog.inf.ed.ac.uk/atate/2024/05/13/firestorm-grids-xml/

Incorporating Pull Request #15  by @nbtafelberg where comments on the rationale for this update are given.

grids.xml file should also be used to replace current http://phoenixviewer.com/app/fsdata/grids.xml 

Recommended approach to add a grid to the Firestorm Grid Manager grid choice list is for the grid to provide an "addgrid" link of form

secondlife:///app/gridmanager/addgrid/http%3A%2F%2Fyour-grid.com%3A8002

Last part of the URL is the URL-encoded grid login URL with : (%3A) / (%2F) and spaces (%20) encoded.

### DIFF
--- a/indra/newview/app_settings/grids.xml
+++ b/indra/newview/app_settings/grids.xml
@@ -220,83 +220,32 @@
         </map>
     <key>goto.theencoreescape.com:8002</key>
         <map>
-        <key>DirectoryFee</key>
-            <string>0</string>
-        <key>LastModified</key>
-            <date>2020-04-24T17:00:50.92Z</date>
-        <key>SendGridInfoToViewerOnLogin</key>
-            <string>true</string>
-        <key>about</key>
-            <string>http://theencoreescape.com/</string>
-        <key>gatekeeper</key>
-            <string>goto.theencoreescape.com:8002</string>
-        <key>gridname</key>
-            <string>The Encore Escape</string>
-        <key>gridnick</key>
-            <string>EncoreEscape</string>
-        <key>help</key>
-            <string>http://theencoreescape.com/</string>
-        <key>helperuri</key>
-            <string>http://money.theencoreescape.com/</string>
+        <key>DEPRECATED</key>
+            <string>TRUE</string>
         <key>login_identifier_types</key>
             <array>
                 <string>agent</string>
                 <string>account</string>
             </array>
-        <key>loginpage</key>
-            <string>http://splash.theencoreescape.com/</string>
-        <key>loginuri</key>
-            <array>
-                <string>http://goto.theencoreescape.com:8002/</string>
-            </array>
+        <key>LastModified</key>
+            <date>2024-05-16T08:00:00.00Z</date>
         <key>name</key>
             <string>goto.theencoreescape.com:8002</string>
-        <key>password</key>
-            <string>http://theencoreescape.com/</string>
-        <key>platform</key>
-            <string>OpenSim</string>
-        <key>register</key>
-            <string>http://theencoreescape.com/</string>
-        <key>search</key>
-            <string>http://search.theencoreescape.com/query.php</string>
-        <key>slurl_base</key>
-            <string>hop://goto.theencoreescape.com:8002/</string>
         </map>
     <key>grid.3rdrockgrid.com:8002</key>
         <map>
-        <key>LastModified</key>
-            <date>2012-08-03T15:32:54.31Z</date>
-        <key>about</key>
-            <string>http://3rdrockgrid.com/</string>
-        <key>gridname</key>
-            <string>3RD Rock Grid</string>
-        <key>gridnick</key>
-            <string>3RG</string>
-        <key>help</key>
-            <string>http://3rdrockgrid.com/</string>
-        <key>helperuri</key>
-            <string>http://grid.3rdrockgrid.com/3rg_money/</string>
+        <key>DEPRECATED</key>
+            <string>TRUE</string>
         <key>login_identifier_types</key>
             <array>
                 <string>agent</string>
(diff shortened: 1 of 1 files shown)


indra/newview/app_settings/grids.xml
```diff
@@ -953,7 +904,7 @@
                 <string>agent</string>
             </array>
         <key>loginpage</key>
-            <string>https://phoenixviewer.com/app/loginV3/</string>
+            <string>http://viewer-login.agni.lindenlab.com</string>
```
Ansariel, line 907: ???

beqjanus: The explanation for this is in the blog Austin linked to this 

> Refinement of handling of Second Life Beta Grid Entry/Splash Page (Aditi)
> 
> Firestorm uses a custom “loginpage” entry/splash page for the main (Agni) grid at https://phoenixviewer.com/app/loginV3/ rather than the usual Linden Lab provided entry at http://viewer-login.agni.lindenlab.com/
> 
> But it also uses this same URL for the Second Life Beta grid (Aditi) which I think is not as helpful as the redirect to a Beta grid specific entry page that is used from http://viewer-login.agni.lindenlab.com when the Second Life Beta grid is selected gives more useful user information when the Second Life Beta grid is selected.

Seems reasonable for the small number that might use it, I am not convinced that we should be removing our own splash though.

PanteraPolnocy: Is there a dedicated Aditi login page? Because even in that address above, I see "agni". Our splash uses the same endpoints as the Lab's - this - one. I see no gain here.

beqjanus: I believe that the agni page presents a different page when the grid is set to aditi, I've not tested that though.  @aiaustin has I presume.

beqjanus:
@aiaustin could you revert the aditi login splash to be the FS one please. That change requires different consideration to the purely OpenSim changes of the grid update.

indra/newview/app_settings/grids.xml
aiaustin: Yes. I also questioned why LL used an agni URL for the aditi splash page. I assumed it was simple redirect and wanted to put that in grids.xml to make the operation and intention clearer.  But I checked and received good help on the SL Discord viewer channel. The agni page has logic to serve the beta page.

The custom Firestorm page for agni works great, but it really is not helpful when selecting the Beta grid. I considered no including that change or just suggestion it in a  comment so the FS devs could decide on what was best. But as I have modified mine own setup for years to make that work better for Beta I included with clear rationale.  Since I was tidying up grids.xml and we may not get another chance I thought I would include that suggestion.

Its easy to change that line back if anyone things the FS splash page works well for the Beta grid user logins.

aiaustin: For documentation this is part of the advice from the Discord Second Life Server "#viewers" channel...

You will find on that the agni page has a whole host of elements of class or id "non-prod-grid-msg" like this one:

```
<div id="non-prod-grid-msg" style="display: none">
  <img class="non-prod-grid-msg-img" src="/assets/uc-hell-yes-b4b2b4c99c72a253d94b2a26adec2642b75ab07a67fe4c9ce3a6331b321750bc.gif"/>
  <h1 id="non-prod-grid-msg-title">Test Grid: </h1>
  <img class="non-prod-grid-msg-img" src="/assets/uc-hell-yes-b4b2b4c99c72a253d94b2a26adec2642b75ab07a67fe4c9ce3a6331b321750bc.gif"/>
  <div id="non-prod-grid-msg-desc"></div>
</div>
```

aiaustin: Of course an option is to provide two separate Firestorm related splash pages (simplest) or use similar web page logic and present a Beta version of the web page (rather obscure).

aiaustin:
FS-with-Beta-Grid-Selected

aiaustin:
> @aiaustin could you revert the aditi login splash to be the FS one please. That change requires different consideration to the purely OpenSim changes of the grid update.

I understand. And no problem. Just lets wait until folks see the explanation added to Ansariel's question before we commit to that shall we?  I nearly did not include the Beta page change as I realised it might need more explanation and discussion.  So its good this got raised here. If a comment can be added into the .xml file it would help anyway but I was not sure if XML comments were permitted.

In any case I need to understand how to make a change to a patch in a PR!

indra/newview/app_settings/grids.xml
aiaustin: > I believe that the agni page presents a different page when the grid is set to aditi, I've not tested that though. @aiaustin has I presume.

FS-with-Beta-Grid-Selected

PanteraPolnocy: I believe that this will suffice: https://phoenixviewer.com/app/loginV3/ vs https://phoenixviewer.com/app/loginV3/?grid=Aditi. Since the viewer already sends the 'grid' in the URL parameter, you can access the updated Aditi splash page even on the current official release of Firestorm. The information text is hardcoded; I assume that the message will not change too often (if at all).

obraz

aiaustin: I agree, that is a great improvement.  I will revert that now.

aiaustin:
Hopefully I did the revert correctly for the Second Life Beta grid to use the custom Firestorm loginpage which now shows Beta grid information. Thanks @PanteraPolnocy

aiaustin:
Improved ways to do this are under consideration. So PR#26 closed as "no need to implement".

PR #27, opened 2024-05-18
FIRE-33613: Allow camera to view negative Z in OpenSim builds

humbletim:
This PR addresses a regression introduced in recent Firestorm releases (7.1.3 and likely earlier) that prevents the camera from being positioned below Z=0 in OpenSim regions. This issue disrupts the ability of users to view avatars and objects situated at negative Z, hindering interaction with underwater builds and other scenarios. 

**Changes:**

- **Introduced `LLViewerRegion::mMinSimHeight` and `LLViewerRegion::getMinSimHeight()`** to track the minimum simulation height advertised by OpenSim regions. 
- **Refined the camera position calculation** in `llagentcamera.cpp` to respect `mMinSimHeight` when applicable, effectively removing the artificial Z=0 floor for OpenSim regions.

This solution leverages existing OpenSim infrastructure by using the `OpenSimExtras.MinSimHeight` value advertised in Simulation Features. If this value is not present, the default behavior is preserved, ensuring compatibility with Second Life.

**See also:**

- [FIRE-33613: [OpenSim] [PBR] Camera cannot be located at negative Z](https://jira.firestormviewer.org/browse/FIRE-33613) Jira bug report
- [metaverse-crossroads/hypergrid-junction#2](https://github.com/metaverse-crossroads/hypergrid-junction/issues/2) for detailed discussion and investigation.

### DIFF
--- a/indra/newview/llagentcamera.cpp
+++ b/indra/newview/llagentcamera.cpp
@@ -2145,10 +2145,21 @@ LLVector3d LLAgentCamera::calcCameraPositionTargetGlobal(BOOL *hit_limit)
     }
 // [/RLVa:KB]
 
+// <FS:humbletim> FIRE-33613: [OpenSim] [PBR] Camera cannot be located at negative Z
+    F32 camera_ground_plane = F_ALMOST_ZERO;
+#ifdef OPENSIM
+    // integrate OpenSimExtras.MinSimHeight into the camera ground plane calculation
+    if (auto regionp = LLWorld::getInstance()->getRegionFromPosGlobal(camera_position_global))
+    {
+      camera_ground_plane += regionp->getMinSimHeight();
+    }
+#endif
+// </FS:humbletim>
+
     // Don't let camera go underground
     F32 camera_min_off_ground = getCameraMinOffGround();
     camera_land_height = LLWorld::getInstance()->resolveLandHeightGlobal(camera_position_global);
-    F32 minZ = llmax(F_ALMOST_ZERO, camera_land_height + camera_min_off_ground);
+    F32 minZ = llmax(camera_ground_plane, camera_land_height + camera_min_off_ground);
     if (camera_position_global.mdV[VZ] < minZ)
     {
         camera_position_global.mdV[VZ] = minZ;
--- a/indra/newview/llviewerregion.cpp
+++ b/indra/newview/llviewerregion.cpp
@@ -675,7 +675,8 @@ LLViewerRegion::LLViewerRegion(const U64 &handle,
 // </FS:Beq>
     // <FS:CR> Aurora Sim
     mWidth(region_width_meters),
-    mWidthScaleFactor(region_width_meters / REGION_WIDTH_METERS) // <FS:Ansariel> FIRE-19563: Scaling for OpenSim VarRegions
+    mWidthScaleFactor(region_width_meters / REGION_WIDTH_METERS), // <FS:Ansariel> FIRE-19563: Scaling for OpenSim VarRegions
+    mMinSimHeight(0.f) // <FS:humbletim> FIRE-33613: [OpenSim] [PBR] Camera cannot be located at negative Z
 {
     // Moved this up... -> mWidth = region_width_meters;
 // </FS:CR>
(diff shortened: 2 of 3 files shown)


beqjanus:
I recently implemented this myself. I'll compare my code with @humbletim's and see if I need to do any more than I have done so far.

humbletim:
note: force push reapplying changes against `phoenix-firestorm/master`; previously they were based on 7.1.7-2nd_preview branch; I *think* whitespace matches expectations now..

indra/newview/llagentcamera.cpp
```diff
@@ -2145,6 +2145,13 @@ LLVector3d LLAgentCamera::calcCameraPositionTargetGlobal(BOOL *hit_limit)
     }
 // [/RLVa:KB]
 
+#ifdef OPENSIM // humbletim:FIRE-33613: Camera cannot be located at negative Z
+    // local variable shadow of the ground plane constant (used in llmax below)
+    F32 F_ALMOST_ZERO = ::F_ALMOST_ZERO;
```
Ansariel, line 2150: I don't really like this. Don't override a global constant with a local variable of the same name.

humbletim: Agreed in principle, though hereit is very intentional and mindfully applied to avoid upstream merge conflicts later...

Could you recommend an alternative strategy that fits in with how such changes to upstream-owned blocks of code are best maintained forward?

humbletim:
> I recently implemented this myself....

ps: I completely understand @Ansariel's point about the local variable. Let's not hold things up! If using your original solution @beqjanus gets a fix across the finish line faster for this release, I'm all for that. We can always circle back and revisit alternative approaches later, especially if there are cleaner ways to tackle the Z-axis issue.

indra/newview/llagentcamera.cpp
beqjanus: From a maintainability perspective, a small reworking to declare F32 ground_plane = F_ALMOST_ZERO; the change the later line while exposing us more to the possibility of future updates, leaves a more readable code. It's not so much the override of a global (though I think we all agree that's generally ill-advised) it is more that F_ALMOST_ZERO is now (potentially) nothing like ZERO. 

I think this works, I've also added explicit "FS guards" around the opensim code change to be consistent.
If you are happy with that, please add a commit to the branch. It would do no harm to squash them into a single commit. and we can merge from there. 

```c++
// <FS:humbletim> FIRE-33613: Camera cannot be located at negative Z
F32 ground_plane = F_ALMOST_ZERO;
#ifdef OPENSIM 
    // modify the ground plancelocal variable shadow of the ground plane constant (used in llmax below)
    if (auto regionp = LLWorld::getInstance()->getRegionFromPosGlobal(camera_position_global)) 
    {
      ground_plane += regionp->getMinSimHeight();
    }
#endif
// </FS:humbletim>
    // Don't let camera go underground
    F32 camera_min_off_ground = getCameraMinOffGround();
    camera_land_height = LLWorld::getInstance()->resolveLandHeightGlobal(camera_position_global);
    F32 minZ = llmax(ground_plane, camera_land_height + camera_min_off_ground); // <FS:humbletim/> FIRE-33613: Camera cannot be located at negative Z
```

humbletim: Sounds good! Strategy implemented, squished to a single commit, and ready to go I think.

<details><summary>Additional thoughts</summary>

Leaning into readability... and as this involves a precedent-setting change (upstream assertion that `Z < 0` falls beyond shared responsibility)... it might also help to pivot around the two competing ground truths concretely (rather than as an offset):
```c++
F32 camera_ground_plane = 0.0f; // Second Life de facto ground plane 
if (OPENSIM) camera_ground_plane = OpenSimExtras.MinSimHeight; // OpenSim de facto ground plane
F32 minZ = llmax(camera_ground_plane + F_ALMOST_ZERO, ...);
```
 
But I digress. This PR fixes the immediate issue though highlights a broader challenge for OpenSim. With Second Life's limitation on negative Z-axis regions, the burden of handling these areas may shift entirely to third-party viewers like Firestorm. Since Second Life itself likely *won't be testing functionality in negative Z-axis regions*, ensuring continued functionality in OpenSim might require significant collaboration between OpenSim and TPV developers.
</details>

PR #28, opened 2024-06-22
fix build target name in docs RelWithDebInfo_open -> RelWithDebInfoFS…

taladar:
I assume whoever wrote that accidentally left out the FS, or at least that one exists and the one without FS does not.


Ansariel:
Thank you for noticing this!

PR #29, opened 2024-06-26
Added camera roll buttons

chanayane:
I added some buttons for camera roll. It's very nice for taking pictures.
Most of the code was adapted from Alchemy viewer.

firestorm-bin_GB9JWvJUri


indra/newview/llagentcamera.cpp
```diff
@@ -937,6 +940,18 @@ void LLAgentCamera::cameraOrbitOver(const F32 angle)
     }
 }
 
+//-----------------------------------------------------------------------------
+// cameraRollOver()
+//-----------------------------------------------------------------------------
+void LLAgentCamera::cameraRollOver(const F32 angle)
+{
+    mRollAngle += fmod(angle, F_TWO_PI);
```
Ansariel, line 948: Should use fmodf instead to avoid type conversions
```suggestion
    mRollAngle += fmodf(angle, F_TWO_PI);
```

Ansariel:
Some remarks from me:
1. The correct attribution is missing in the commit
2. The code should be tagged/marked as FS-specific code - see other places where the code diverges from LL code
3. What about the empty space between the buttons? Seems a bit of wasted space.
4. Is it intentional they should only be added to the default camera floater but not the smaller version (floater_fs_camera_small.xml)?
5. Would have been nice to add the changes to the AnsaStorm-specific skin

chanayane:
> Some remarks from me:
> 
> 1. The correct attribution is missing in the commit
> 2. The code should be tagged/marked as FS-specific code - see other places where the code diverges from LL code
> 3. What about the empty space between the buttons? Seems a bit of wasted space.
> 4. Is it intentional they should only be added to the default camera floater but not the smaller version (floater_fs_camera_small.xml)?
> 5. Would have been nice to add the changes to the AnsaStorm-specific skin

My apologies, this is the first time I contribute to Firestorm.
1. What do you mean by correct attribution? Should I change the commit message and say it comes from Alchemy viewer? And/or something else?
2. I'll mark the code
3. About the UI, I am listening to suggestions. The way they did it in Alchemy, they added the buttons on the bottom of the camera orbiting button. But it doesn't work well for Firestorm because that panel is resizable.
4. When we agree on point 3, I'll add it to the smaller version
5. and AnsaStorm as well.

Thank you for reviewing my PR and I'll rework on it.

Ansariel:
> > Some remarks from me:
> > 
> > 1. The correct attribution is missing in the commit
> > 2. The code should be tagged/marked as FS-specific code - see other places where the code diverges from LL code
> > 3. What about the empty space between the buttons? Seems a bit of wasted space.
> > 4. Is it intentional they should only be added to the default camera floater but not the smaller version (floater_fs_camera_small.xml)?
> > 5. Would have been nice to add the changes to the AnsaStorm-specific skin
> 
> My apologies, this is the first time I contribute to Firestorm.
> 
>     1. What do you mean by correct attribution? Should I change the commit message and say it comes from Alchemy viewer? And/or something else?
> 
>     2. I'll mark the code
> 
>     3. About the UI, I am listening to suggestions. The way they did it in Alchemy, they added the buttons on the bottom of the camera orbiting button. But it doesn't work well for Firestorm because that panel is resizable.
> 
>     4. When we agree on point 3, I'll add it to the smaller version
> 
>     5. and AnsaStorm as well.
> 
> 
> Thank you for reviewing my PR and I'll rework on it.

Re 1: If you take code from other viewers, you need to attribute the original author and source properly according to the license used.

Re 3: No idea. I just noticed it will make the floater wider for 2 small buttons and there is a lot of unused space. Maybe there is a way to optimize it.

chanayane:
Okay, I reworked my commit, hopefully I understood what you wanted me to do, but if not, let me know.

1. Changed the commit message to say that the code mostly comes from Alchemy viewer, providing info on the license of Alchemy source code
2. Marked all my modifications. I took example on how you marked your changes.
3. Changed the UI so that the buttons are located under the camera rotate stick for Firestorm normal, small and Ansastorm skin
4. Applied your code suggestion about fmod and fmodf

Here is what it looks like:

firestorm_camera_floater_normal
firestorm_camera_floater_small
firestorm_camera_floater_ansastorm

Ansariel:
The layout looks good to me. Not sure if we need to explicitly mention the original author - I'll let @beqjanus decide about that.

PR #30, opened 2024-06-29
Add option to avoid touching attachments when shift clicking avatar

makidoll:
I have lots of attachments on my avatar and it's become increasingly difficult to click on my avatar to focus and move.

This PR adds an option that lets you avoid touching attachments whilst pressing shift when clicking your avatar.

The only issue I've encountered is that hud attachments stop being touchable when holding shift.

The cursor also still indicates that attachments are clickable.

image

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -22255,6 +22255,17 @@ Change of this parameter will affect the layout of buttons in notification toast
         <key>Value</key>
         <integer>1</integer>
     </map>
+    <key>FSShiftClickAvoidsAvatarAttachments</key>
+    <map>
+        <key>Comment</key>
+        <string>If true, shift clicking on avatar will avoid touching attachments</string>
+        <key>Persist</key>
+        <integer>1</integer>
+        <key>Type</key>
+        <string>Boolean</string>
+        <key>Value</key>
+        <integer>0</integer>
+    </map>
     <key>FSShowInterfaceInMouselook</key>
     <map>
         <key>Comment</key>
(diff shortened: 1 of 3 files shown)


beqjanus:
Hi @makidoll thanks for the PR and sorry for the long wait. 
I just tested this and it appears to prevent shift-click on hud attachments. I am not sure how common the use of shift-click in HUDs is but I can bet someone who wants this feature enabled will also have a HUD that stops working

I suspect it is just a case of excluding HUDattachments when doing the pick, but as I am not sure I have the right avatar setup to test this adequately I'd prefer it if you could have a look please.

Also, please add the //<FS> comment blocks around code that is added. 

typically If you are modifying LL code  We need the LL code preserved in a comment.
For example:

```c++
int buggy_code = true; 
LL_WARN() << "This code is buggy" << LL_ENDL;
```
would become (You can use \<FS:Maki\> or just plain \<FS\>)
```c++
// <FS:Name> Fix the buggy code 
// int buggy_code = TRUE; 
// LL_WARN() << "This code is buggy" << LL_ENDL;
bool fixed_code = true;
LL_DEBUG() << "I fixed this" << LL_ENDL;
// </FS:Name>
```

If you are simply adding new code then the same rules apply but there is nothing to comment out. 
This is done so that when LL update the original code we can see what the original code was doing, what their changes do and how that relates to the changes that we applied on top.

beqjanus:
I've added a new guideline page https://github.com/FirestormViewer/phoenix-firestorm/blob/master/FS_PR_GUIDELINES.md that might explain how these work a bit better

beqjanus:
Hi @makidoll 
Just checking whether you saw the previous comment and whether you intend to make the requested changes.

beqjanus:
Closing this as it has become inactive. @makidoll feel free to reopen if you come back to this

PR #31, opened 2024-07-09
Raised the 512x512 limitation for uploading snapshots to inventory to 2048x2048

chanayane:
I don't know if there is a reason why snapshots to inventory have been blocked to 512x512 until now. If there are no reasons, then here is a pull request that allow to upload snapshots to inventory up to 2048x2048.
The cost of the upload is updated according to the area of the snapshot.

### DIFF
--- a/indra/newview/llagentbenefits.cpp
+++ b/indra/newview/llagentbenefits.cpp
@@ -267,6 +267,25 @@ S32 LLAgentBenefits::getTextureUploadCost(const LLImageBase* tex) const
     return getTextureUploadCost();
 }
 
+// <FS:Chanayane> 2048x2048 snapshots upload to inventory
+S32 LLAgentBenefits::getTextureUploadCost(S32 w, S32 h) const
+{
+    if (w > 0 && h > 0)
+    {
+        S32 area = w * h;
+        if (area >= MIN_2K_TEXTURE_AREA)
+        {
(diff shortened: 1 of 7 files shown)


Ansariel:
First of all: Thank you for the contribution!

This is interesting and it made me wonder why snapshots were always limited to 512x512px while you could technically also upload to a maximum of 1024x1024px. I wonder if there is an actual reason for this or nobody - especially at LL - thought of changing this. It actually might be a good idea to submit this to Linden Lab directly and see if they accept it or deliver an explanation as to why this shouldn't be done.

chanayane:
> First of all: Thank you for the contribution!
> 
> This is interesting and it made me wonder why snapshots were always limited to 512x512px while you could technically also upload to a maximum of 1024x1024px. I wonder if there is an actual reason for this or nobody - especially at LL - thought of changing this. It actually might be a good idea to submit this to Linden Lab directly and see if they accept it or deliver an explanation as to why this shouldn't be done.

I made a PR to the official viewer. We'll see what they'll say about it.

https://github.com/secondlife/viewer/pull/2025

chanayane:
My PR got merged in SL viewer repository. Should I close that PR as it is now redundant?

Ansariel:
Leave it open - I will merge it tomorrow or so.

PR #32, opened 2024-07-11
annotate gcc 13 dangling reference to temporary errors so they are ignored

taladar:
Mostly errors like

```
phoenix-firestorm/indra/newview/fsfloaterimcontainer.cpp: In static member function ‘static void FSFloaterIMContainer::reloadEmptyFloaters()’:
phoenix-firestorm/indra/newview/fsfloaterimcontainer.cpp:433:42: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  433 |     LLFloaterReg::const_instance_list_t& inst_list = LLFloaterReg::getFloaterList("fs_impanel");
      |                                          ^~~~~~~~~
phoenix-firestorm/indra/newview/fsfloaterimcontainer.cpp:433:82: note: the temporary was destroyed at the end of the full expression ‘LLFloaterReg::getFloaterList(std::__cxx11::basic_string<char>(((const char*)"fs_impanel"), std::allocator<char>()))’
  433 |     LLFloaterReg::const_instance_list_t& inst_list = LLFloaterReg::getFloaterList("fs_impanel");
      |                                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
```

### DIFF
--- a/indra/llui/llfloaterreg.h
+++ b/indra/llui/llfloaterreg.h
@@ -123,7 +123,18 @@ public:
     static bool destroyInstance(const std::string& name, const LLSD& key = LLSD());
 
     // Iterators
+    // <FS> ignore dangling reference false positives in gcc13
+#if defined(__GNUC__) && (__GNUC__ >= 13)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdangling-reference"
+#endif
+    // </FS>
     static const_instance_list_t& getFloaterList(const std::string& name);
+    // <FS> ignore dangling reference false positives in gcc13
+#if defined(__GNUC__) && (__GNUC__ >= 13)
+#pragma GCC diagnostic pop
+#endif
+    // </FS>
 
     // Visibility Management
 // [RLVa:KB] - Checked: 2012-02-07 (RLVa-1.4.5) | Added: RLVa-1.4.5
--- a/indra/newview/llpathfindingnavmesh.cpp
+++ b/indra/newview/llpathfindingnavmesh.cpp
@@ -139,7 +139,18 @@ void LLPathfindingNavMesh::handleNavMeshResult(const LLSD &pContent, U32 pNavMes
         ENavMeshRequestStatus status;
         if ( pContent.has(NAVMESH_DATA_FIELD) )
         {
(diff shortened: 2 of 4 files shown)


Ansariel:
These warning produced by GCC are actually false positives. Both paths in LLFloaterReg::getFloaterList return a reference to a static class variable and not a temporary variable constructed inside getFloaterList. The warning should be disabled specifically for this method and GCC 14

taladar:
Okay, I checked how to do that and now changed the actual code back and surrounded it with annotations to locally suppress the warning for the affected lines only.

Ansariel:
I was rather thinking if it possible to set the pragmas only around LLFloaterReg::getFloaterList instead of sprinkling all over the place. Also you would need to encapsulate the pragma preprocessor instruction in an ifdef like `#if defined(__GNUC__) && (__GNUC__ >= 14)` or Visual Studio and clang will complain about the unknown pragma.

taladar:
Okay, I now found a way to annotate the function instead of the call site and also wrapped the pragmas (though with version 13 when the warning was first introduced).

beqjanus:
@taladar Can I be a pain and ask that you add the comment guards around the blocks of code that need it. That would be the floaterreg and pathfinding modules but not the RLV as that is not LL maintained code. 
I've created a PR Guidelines doc that (hopefully) explains how these work and why we use them. 
https://github.com/FirestormViewer/phoenix-firestorm/blob/master/FS_PR_GUIDELINES.md

taladar:
@beqjanus Like that?

beqjanus:
@taladar Thats in the right places. In general, they should have a description like `// <FS> ignore dangling reference false positives in gcc13` so that in 12 months time we know why they were put in.

taladar:
Okay, I updated it to include a comment, I was not sure what to use in that location since it is usually the place for the ticket and there is none in this case.

beqjanus:
That looks good. Thanks for the PR. Agreed that that is where the ticket would go typically. Ideally you could raise a ticket of course, the main reason for the ticket number is that it allows us to close them and refer to them in release notes etc. But in this case the most important thing is to have the comment so we know why we have all this "clutter" in a few months time when we've all forgotten :-)

PR #33, opened 2024-07-12
Fix: Stop textures from being set as inactive if still on a face.

TommyTheTerrible:
Textures were being set to Inactive even if still in scene, causing them to be deleted and re-decoded in a loop.


Ansariel:
Thank you for your contribution! However, this part has significantly changed in LL's latest version in their "develop"-branch, so this change would only be short-lived. I would suggest you also - or actually - try to contribute the fix directly to LL.

taladar:
If this works as described this could be the fix for https://jira.firestormviewer.org/projects/FIRE/issues/FIRE-34034

TommyTheTerrible:
I read the LL Viewer code and seems to still have the issue, but they can probably just use a BOOL to stop it in the same function.

For example, BOOL starts as FALSE outside the channel loop, set to TRUE if a face is found then IF BANG BOOL the setInactive.

I'll submit a PR to LL, but anyone that needs the fix sooner than later should use this patch.

TommyTheTerrible:
> If this works as described this could be the fix for https://jira.firestormviewer.org/projects/FIRE/issues/FIRE-34034

That looks like the problem. The code prior to patch will go into a loop that will often reload the FastCache raw (if FastCache is enabled) then go down the discards again with the texture file, decoding over and over again.

So, it would go from crisp (discard 0-1) then blurry (discard 4 if FastCache or 5 if reloaded completely) or even gray if decode cannot commence for some reason then start getting crisp again in a loop.

taladar:
Well, I am affected on Linux, I suspected Windows was affected as well but haven't seen any reports yet but purely based on the fact that there was nothing in that whole logic that was platform or hardware specific it was unlikely that it would affect just two of three platforms.

MikeUhuru:
I compiled in Linux *with* the proposed code and with KDU, as I am a license holder, and now a specific spot where I see discarding occur, remained blurred *forever* after a short while. I recompiled with the original code and then it was gone.

taladar:
For me this seems to solve the issue of textures staying gray, the download queue actually settles after a while unlike before where it kept growing due to textures being thrown away.

TommyTheTerrible:
> I compiled in Linux _with_ the proposed code and with KDU, as I am a license holder, and now a specific spot where I see discarding occur, remained blurred _forever_ after a short while. I recompiled with the original code and then it was gone.

I think this is a separate issue. After experimenting trying to fix the OpenJPEG partial decode issues, I believe the two decoding libraries handle partial decode failures differently.

I think the OpenJPEG continues with the partial decode and returns an image that is partially decoded that will have color components unfinished (which will look partially gray).

But I believe the KDU will only return the last good decode with all components finished, so you are seeing a texture from KDU that is a discard level higher than requested which is often blurrier.

So, I think your blurry texture issue is being caused by the guestimation calculations for the partial file decoding are bad for those particular textures. You can test if this is true by finding the line in lltexturefetch.cpp that assigns the size using LLImageJ2C::calcDataSizeJ2C.

By default the calculation uses a 0.85 factor, assuming compression, and I have tried it at 1 still with issues, but if you remove the calculation entirely and just set it to MAX_IMAGE_DATA_SIZE, then things should decode properly.

But keep in mind that this will slow the decoding process down because it will be feeding the decoder more file than necessary.

If you want to keep things speedy, you can try experimenting with the factor by adding a fifth variable to the LLImageJ2C::calcDataSizeJ2C or perhaps see if there is a better way to calculate the texture size per discard.

TommyTheTerrible:
For anyone still following here and having gray or blurry textures, I was able to update the file size calculation to something more reasonable and seems to be working well.

New code in this commit:

https://github.com/TommyTheTerrible/Firestorm-Vulkan/commit/08e59afd63d9710250aebe166a64d1b418afabcf

chanayane:
> For anyone still following here and having gray or blurry textures, I was able to update the file size calculation to something more reasonable and seems to be working well.
> 
> New code in this commit:
> 
> [TommyTheTerrible@08e59af](https://github.com/TommyTheTerrible/Firestorm-Vulkan/commit/08e59afd63d9710250aebe166a64d1b418afabcf)

I'm interested to try your fixes, could you resolve the conflicts? :)

TommyTheTerrible:
I'm sorry I made a mess of this PR. I will try to be more diligent when creating patch branches in the future.

PR has been resubmitted as a mergeable solution.

https://github.com/FirestormViewer/phoenix-firestorm/pull/35

chanayane:
Mmh it still says that the branch has conflicts ^^

chanayane:
Oh I see you made a new PR

PR #34, opened 2024-07-13
Fixes bad upload quality issue with OpenJPEG

chanayane:
I recently noticed that using my self-compiled viewer, all my uploads looked very bad, with a lot of artifacts. The greater the dimensions of the texture, the ugliest it looked, like, over-compressed JPEGs.

I do not have KDU so I have to use the OpenJPEG implementation in my viewer. This affected the uploading of textures and the saving of snapshots to inventory. I think other viewers might be affected as well, so maybe that's not specific to Firestorm.

I discovered that the issue was linked to the OpenJPEG implementation, the maximum size for all encoded images was set to 32KB. So 1024x1024 and especially 2048x2048 textures looked really bad. 

The bad line:
```parameters.max_cs_size = (1 << 15);```
This was defined in the constructor of the JPEG2KEncode class and applied to all pictures regardless of their dimensions.
The documentations says ```max_cs_size: Maximum size (in bytes) for the whole codestream.
If == 0, codestream size limitation is not considered. If it does not comply with tcp_rates, max_cs_size prevails and a warning is issued.```

My fix is:

1. Computing the number of layers depending on image surface, much like what is done in the KDU implementation. So I moved the definition of num_layers and tcp_rates values to the encode method.
2. Defining the rates according to the number of layers, the last one gets set to 1 / DEFAULT_COMPRESSION_RATE instead of the fixed 30.0f to avoid reducing quality of the topmost layer, and each previous layer multiplies its subsequent layer with a multiplier (I used 15, 4, 2... to be close to the previous defaults of 120, 480, 960 and 1920)
3. Computing max_cs_size: for each layer, multiply the surface by the number of components and by 1/layer rate and add the result to max_cs_size
 
So a big texture would be allowed more size and more layers than a little one.

After the fix, the quality of the uploaded textures has improved greatly. In some cases I can still see some very minor compression artefacts and blurriness but they are not too noticeable.

### DIFF
--- a/indra/llimagej2coj/llimagej2coj.cpp
+++ b/indra/llimagej2coj/llimagej2coj.cpp
@@ -447,18 +447,28 @@ public:
             parameters.irreversible = 0; // should be the default, but, just in case
             // </FS:Chanayane>
             parameters.tcp_numlayers = 1;
-            parameters.tcp_rates[0] = 1.0f;
+            // <FS:Chanayane> Fixes bad upload quality issue with OpenJPEG
+            //parameters.tcp_rates[0] = 1.0f;
+            /* documentation is wrong, should be 0.0f for lossless! 
+               see https://github.com/uclouvain/openjpeg/blob/39e8c50a2f9bdcf36810ee3d41bcbf1cc78968ae/src/lib/openjp2/j2k.c#L7755
+            */
+            parameters.tcp_rates[0] = 0.0f;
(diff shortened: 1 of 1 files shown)


chanayane:
I think this should be in the official viewer so that all the other viewers can benefit from the fix.
I made a PR on the official viewer https://github.com/secondlife/viewer/pull/2054
I'll close the Firestorm PR.

PR #35, opened 2024-07-15
Fix: Texture Fixes to Stop Recode Loop and Bad Decodes (Mergeable)

TommyTheTerrible:
I apologize for the last PR, I did not realize I had not made a patch branch.

I refactored the code to be more similar to what Linden Labs accepted as solution for the unnecessary texture unloading and included the updated file size calculations to help the decodes receive a more reliable amount of file.

Edit: This is a resubmission of PR 33: https://github.com/FirestormViewer/phoenix-firestorm/pull/33

### DIFF
--- a/indra/llimage/llimagej2c.cpp
+++ b/indra/llimage/llimagej2c.cpp
@@ -265,11 +265,12 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
     S32 surface = w*h;
     S32 s = 64*64;    
     S32 precision = 8; // assumed bitrate per component channel, might change in future for HDR support
-    S32 totalbytes = (S32)(s * comp * precision * rate); // total assumed bits in pyramid, first level computed before loop
+    S32 max_components = 4; // assume file has the maximum number of components, three for color and one for alpha.
+    S32 totalbytes = (S32)(s * max_components * precision * rate); // total assumed bits in pyramid, first level computed before loop
     while (surface > s)
     {
         if (nb_layers <= (5 - discard_level))
-            totalbytes += (S32)(s * comp * precision * rate);
+            totalbytes += (S32)(s * max_components * precision * rate);
         nb_layers++;
         s *= 4;
     }
--- a/indra/llimage/llimagej2c.cpp
+++ b/indra/llimage/llimagej2c.cpp
@@ -263,13 +263,13 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
     // Estimate the number of layers. This is consistent with what's done for j2c encoding in LLImageJ2CKDU::encodeImpl().
     S32 nb_layers = 1;
     S32 surface = w*h;
-    S32 s = 64*64;
+    S32 s = 64*64;    
     S8  precision = 8; // assumed bitrate per component channel
-    S32 totalbytes = 0; // total assumed bits in pyramid
+    S32 totalbytes = (s * comp * (S32) precision) * rate; // total assumed bits in pyramid, first level computed before loop
     while (surface > s)
     {
-        if (nb_layers <= (6 - discard_level))
-        totalbytes += (s * comp * (S32) precision) * rate;
+        if (nb_layers <= (5 - discard_level))
+            totalbytes += (s * comp * (S32) precision) * rate;
         nb_layers++;
         s *= 4;
     }
@@ -290,9 +290,9 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
     S32 new_bytes = (S32) (sqrt((F32)(w*h))*(F32)(comp)*rate*1000.f/layer_factor);
     S32 old_bytes = (S32)((F32)(w*h*comp)*rate);
     bytes = (LLImage::useNewByteRange() && (new_bytes < old_bytes) ? new_bytes : old_bytes);
-    bytes = llmax(totalbytes, calcHeaderSizeJ2C());
-    //LL_WARNS() << "Calculate w-h-c-d-p " << w << "-" << h << "-" << comp << "-" << discard_level << "-" << (S32)precision
-    //           << " Test : " << totalbytes << " new : " << new_bytes << " old : " << old_bytes << LL_ENDL;
+    bytes = llmax((S32)totalbytes, calcHeaderSizeJ2C());
+    LL_WARNS() << "Calculate w-h-c-d-p " << w << "-" << h << "-" << comp << "-" << discard_level << "-" << (S32)precision
+               << " Pyramid: " << (S32)totalbytes << " LayerFactored: " << new_bytes << " WJCR: " << old_bytes << LL_ENDL;
     return bytes;
 }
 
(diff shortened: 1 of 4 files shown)


beqjanus:
Thanks Tommy, I'll have a look. Given that these are already upstream in develop, we may just pull in the upstream commit as it will make merges less messy later.

beqjanus:
Given that we explicitly add the headersize hackery at line 279, the check at 293 is now redundant as totalBytes can never be less than calcHeaderSize.

TommyTheTerrible:
I left that in just in case someone wanted to test the old two methods.

The code itself should be cleaned up at some point if the new pyramid walking calculation is more beneficial.

But I guess that can be done in LL's develop after it passes testing and reaches an official branch.

I am wondering if the 600 bytes for header is not enough because I see a lot of tiny files get decode errors but it may be because they are trying to be decoded using a discard too high for their size.  I'm investigating this now, may require changing mMinDesiredDiscardLevel to 0 for files smaller than 64x64 in llviewertexture.cpp somewhere.

beqjanus:
Sounds good. I was looking at the 600 bytes, too. It has an ancient comment about changing it once textures are not served by regions. I presume that is related to the expected MTU from the CDN versus the long-haul for most of us, but in any case, I am not sure I understand why it would change the actual useful header size of as texture. The idea is to get the smallest amount that gives the necessary data, but that should not depend on where the thing is being served from. I suspect that the thinking is wrong on that. If you are measuring the first packet/header explicitly then that will give a better idea.

There was a person named "Wondering Fairelander" in the preview chat the other day, I think that was you, you had some thoughts on the multi-threaded GL.

Also I've added [some guidelines to the front-page](https://github.com/FirestormViewer/phoenix-firestorm/blob/master/FS_PR_GUIDELINES.md) that explain how out comment markers work. If you are committing upstream there is no need but for sending to us it would be good to add those it makes merging a lot easier. 

Thank you, this is good stuff.

TommyTheTerrible:
I agree, it may have made sense in the distance past for some reason but I think the maximum possible header for a J2C file should be used instead. I have not figured out what that is yet and things seem to be running okay for 99% of the textures being decoded so I guess not a big deal for now.

Wondering Fairelander is my test account that I use to avoid sending online/offline messages a lot to friends on my main account as I am testing builds. I'm sorry if I caused any issue, I don't know where to chat about this stuff in-world and sometimes I think I irk the group moderators.

As far as that particular topic, about the multithreading texture changes, seems to be the raw cache reset flooding in situations where the scene changes drastically (teleporting out of a complex scene or reducing your draw distance drastically in a complex scene). On slower RAM it does not seem to be an issue, since I am fine on my DDR4 but a friend with the same exact video card but DDR5 RAM will crash in those situations.

I think it can be solved by checking the mTextureState before doing the raw cache resets but I haven't finished testing myself, much less had them test on the DDR5.

I was hoping we can keep this PR open for self-compilers to see the changes but I will be submitting to LL as well, so I don't know how to handle that with your guidelines. Should I submit here as well with the SMGL-style comments?

Ansariel:
I cleaned the change to llimagej2c.cpp in LL's development branch up already.

As for the changes to the texturelist: I don't think this change is correct and Henri Beauchamp thinks the same. See my comment for https://github.com/secondlife/viewer/commit/27a2531c5bf375ed208d8fd5764a8797f08ce831

TommyTheTerrible:
I'm sorry I did not see your comment on the commit, Ansariel. I thought Github was suppose to notify us of that sort of thing? I responded there for continuity.

I found an issue with my pyramid calculation, where the components value provided does not account for the possibility of an alpha channel. Unfortunately, the needs_aux sent to LLTextureFetch::createRequest is only set to true if the face has an alpha mask option enabled. So, the calculation itself needs to assume 4 components, otherwise there are decoding issues for the color channel as well as the aux channel later.

I will submit an update to the code in a few moments.

beqjanus (commented):
Silly as this might sound but I have a nasty feeling that max_components is actually 5 because there is this weird extra mask channel on some bakes (a bump channel on the eye bake I think?). I could be wrong @Ansariel will likely remember better than I do. What I don't know is whether it ever applies to fetched textures or not and whether it makes any difference here.

Ansariel:
As mentioned on the commit for the LL viewer: Have you tried this instead? Because simply preventing a texture from being set inactive just because it happens to be in the texture list is likely the wrong way:
```
        else if (imagep->isInactive())
        {
            if (imagep->getBoundRecently())
            {
                imagep->forceActive();
                imagep->getLastReferencedTimer()->reset();
            }
            else
            {
                if (imagep->getLastReferencedTimer()->getElapsedTimeF32() > max_inactive_time)
                {
                    imagep->setDeletionCandidate();
                }
                return;
            }
        }
        else
        {
            imagep->getLastReferencedTimer()->reset();

            //reset texture state.
            if(!onFace)
                imagep->setInactive();
        }
```

TommyTheTerrible:
@beqjanus, I think the baked stuff is safe because it always uses discard 0 and the max file size.

I responded to the commit thread with the suggestion of just making the setInactive accept a bool and swap the state, that way we can just move that line of code to the top.

I can write up the code if you want, should be easier to read than the nested if statements and I personally avoid branching if I can for performance reasons, but I'm happy my textures are behaving and okay with whatever gets the job done.

TommyTheTerrible:
Added code to allow the inactivity grace period to work again.

I made the assumption that once a texture is a delete candidate there is no going back, which I don't think is fair because everyone deserves a second chance. Perhaps allowing up to the DELETED state?

Ansariel:
The "on face" idea is generally wrong because it will prevent textures from being removed from memory if it isn't rendered at all and - that's how it used to work - get removed from memory if a certain threshold of memory load was reached. I explained it in detail on the actual commit for the LL viewer.

Ansariel:
I did some testing with current master and the development branch, and I think the correct solution would be to reinstate a threshold in LLViewerFetchedTexture::destroyTexture() so textures only get destroyed after a certain amount of VRAM is used. There is no additional change regarding the last referenced timer required. What it also does is - if VRAM didn't reach the threshold yet - it stops this constant wild texture fetch activity: The texture fetch queue is empty.

TommyTheTerrible:
I do not agree on reverting to the older destroyTexture code. I wrote a lengthier reply in the commit thread.

I know I am new here and should respect seniority but I think we should be making adjustments towards a goal and not knee-jerking back to older code that may have been changed for a reason.

Ansariel:
Sorry to disagree, but there is nothing wrong with caching textures and removing them if the space they are occupying is required otherwise. What you do not want is to unnecessarily decode images from raw again if you are caming around for instance.

The older code was clearly not removed because it was bad performance-wise - in that case the whole "marked for deletion" concept would have been gone. The particular part in `LLViewerFetchedTexture::destroyTexture()` got removed because it resembled around the artificial texture memory limit, and with that being removed the metrics around it got removed as well. That caused exactly the problem you are now trying to solve by some other solution without calling it "caching".

It's also not - as you described at the commit itself - that thousands of textures would be removed in a single frame and cause a huge frame stutter: `LLViewerFetchedTexture::destroyTexture()` would only be called for as many textures per frame that fit into a 2-5ms time slot. But like I said: This is currently done anyway already, actually for even more textures at once since no caching is involved.

Ansariel:
Suggestion: We should wait until Runitai is back (from vacation?) and then see what he has to say about this, since he is mostly responsible for the changes that happened to the texture pipeline.

PR #40, opened 2024-08-13
[FIRE-34403] Recurse and fetch entire inventory at login for OpenSim to fix inventory issues

Hecklezz:
[[FIRE-34403]](https://jira.firestormviewer.org/browse/FIRE-34403)

Fixes the problems mentioned in the JIRA issue.
(As talked about in the JIRA issue, an alternative and more direct approach to fixing the issues mentioned might be more appropriate than what I propose in this PR)

### DIFF
--- a/indra/newview/llinventorymodelbackgroundfetch.cpp
+++ b/indra/newview/llinventorymodelbackgroundfetch.cpp
@@ -205,7 +205,7 @@ LLInventoryModelBackgroundFetch::LLInventoryModelBackgroundFetch():
 
 LLInventoryModelBackgroundFetch::~LLInventoryModelBackgroundFetch()
 {
-    gIdleCallbacks.deleteFunction(&LLInventoryModelBackgroundFetch::backgroundFetchCB, NULL);
+    gIdleCallbacks.deleteFunction(&LLInventoryModelBackgroundFetch::backgroundFetchCB, nullptr);
 }
 
 bool LLInventoryModelBackgroundFetch::isBulkFetchProcessingComplete() const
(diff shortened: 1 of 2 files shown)


Hecklezz:
Updated the PR to reflect the exact change wanted, as well as a general cleanup of the code thanks to @Ansariel along with extras I did myself.
Line 1300 is where the actual fix is made, everything else is refactoring and some styling fixes.

PS: There are a lot more code styling errors in this file that I spotted that have to do with inconsistencies of pointer and reference alignment. Sometimes they are left, sometimes right and sometimes centre. There are also a few other styling errors throughout as well. If you want me to clean these up while I'm at it, I can, but otherwise this PR contains just the important and relevant ones (such as using emplace everywhere instead of push which will save on unnecessary copies in a number of places).

Hecklezz:
> I already committed most these changes upstream to LL's develop branch and will come in the future. No need to do an extra commit for these now.
I know, I also have a PR up over at LL's viewer too which has a large number of things you missed that aren't present in this PR- 
https://github.com/secondlife/viewer/pull/2476

Ansariel:
> I know but it also has the explicit change that Ubit wanted on line 1300. 

That is what Ubit wanted and I implemented what he wanted to make it work in OS (which is basically your original change just slightly different). I don't want to keep it for SL the way it is, so that code should only be called if `LLViewerInventoryCategory::VERSION_UNKNOWN != cat->getVersion()` for SL.

Hecklezz:
My original PR already did EXACTLY what your PR did but it was never merged because they wanted the else removed entirely.
Ubit wanted the else entirely removed because there is no reason to only check for SL.

Ansariel:
Ubit may speak for OpenSim, but not for SL, so I prefer doing exactly the same thing in SL as the LL viewer does.

Hecklezz:
Well I guess it is up to you, but there is a good chance that in SL it would exhibit the same issue as OpenSim does if AISv3 is disabled (which it probably almost never is), in that the entire inventory won't be fetched properly like it does when AISv3 is enabled.

Ansariel:
There is no UDP inventory fallback in SL - lots of the UDP methods have been removed from the viewer and on the server side as well.

Ansariel:
I'm closing this PR since we will get those changes in `develop` anyway and I prefer sticking with my OpenSim fix variant.

PR #41, opened 2024-08-14
FIRE-31508: check folder limit before attaching all the things in a f…

AngeldarkRaymaker:
Hi there.

Picked a case that's gotten my goat several times: accidentally attaching all the things in the root of all my clothes or hair...

It appears to be a a regression from "ReplaceWornItemsOnly", no case ID, and I didn't look it up.

To solve it, I elected to refactor rather than duplicate code, is that okay? Do you prefer keeping original?

### DIFF
--- a/indra/newview/llinventorybridge.cpp
+++ b/indra/newview/llinventorybridge.cpp
@@ -3680,6 +3680,11 @@ void LLFolderBridge::performAction(LLInventoryModel* model, std::string action)
 // <FS:TT> Patch: ReplaceWornItemsOnly
     else if ("replaceitems" == action)
     {
+        //<FS:AR> FIRE-31508: check folder limit
+        if (modifyOutfitExceedsWearFolderLimit())
+            return;
+        //</FS:AR> FIRE-31508
+
         LLInventoryModel* model = getInventoryModel();
         if(!model) return;
         LLViewerInventoryCategory* cat = getCategory();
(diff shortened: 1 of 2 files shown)


indra/newview/llinventorybridge.cpp
```diff
@@ -3680,6 +3680,11 @@ void LLFolderBridge::performAction(LLInventoryModel* model, std::string action)
 // <FS:TT> Patch: ReplaceWornItemsOnly
     else if ("replaceitems" == action)
     {
+        //<FS:AR> FIRE-31508: check folder limit
+        if (modifyOutfitExceedsWearFolderLimit())
```
AngeldarkRaymaker, line 3684: Expand up once, and you'll see modifyOutfit(bool).
In there the check/nag is done, so it seemed to be the easier thing to do: use what is already there.

indra/newview/llinventorybridge.cpp
```diff
@@ -5368,32 +5373,65 @@ void LLFolderBridge::createWearable(LLFolderBridge* bridge, LLWearableType::ETyp
...
 
-void LLFolderBridge::modifyOutfit(BOOL append)
+//<FS:AR> FIRE-31508: refactored from void LLFolderBridge::modifyOutfit(BOOL append)
+bool LLFolderBridge::modifyOutfitExceedsWearFolderLimit()
 {
-    LLInventoryModel* model = getInventoryModel();
-    if(!model) return;
-    LLViewerInventoryCategory* cat = getCategory();
-    if(!cat) return;
+    LLInventoryModel *model = getInventoryModel();
```
AngeldarkRaymaker, line 5379: I don't know what this is, doesn't appear used... 
I have no idea what state it creates (if any). It's in the old code, so I carried it around.

Ansariel: This is either obsolete or `model` should be used instead of `gInventoryModel` to collect the category contents a few line further down.

AngeldarkRaymaker: `model` appears vestigial...
`getCatagory()` itself performs a `getInventoryModel()` for its interrogation... it might be a hangover?

`getCategory()` will yield null if `getInventoryModel()` yields null, so appears safe enough to comment out.

Ansariel:
Looks like it needed another minute after resolving the conversations.

PR #43, opened 2024-09-17
Add key bindings for Camera Roll

chanayane:
Adds key bindings for the new Camera Roll feature that got accepted a few weeks ago. I think this will make the camera roll feature more useful for people who use keyboard shortcuts to control the camera.

By default, I chose the following keys:
- R (roll left)
- T (roll right)

I couldn't find any other use for these keys so I think it's alright without any modifier, and also convenient. Easy to remember as R is for Roll.

You might have to define the keys yourself or reset key bindings to defaults after an update for this to work, at least I had to do it on my computer.

You can see the new key bindings in Preferences > Controls > When in third person or When sitting, under the Camera section.

special-FirestormOS-AyaneStorm_HoYzZEMLY5

### DIFF
--- a/indra/newview/app_settings/key_bindings.xml
+++ b/indra/newview/app_settings/key_bindings.xml
@@ -35,6 +35,10 @@
     <binding key="E" mask="NONE" command="jump"/>
     <binding key="C" mask="NONE" command="push_down"/>
 
+<!-- <FS:Chanayane> Camera roll key bindings -->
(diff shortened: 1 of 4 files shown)


beqjanus:
Thanks for this. I  might ponder the choice of controls. It would seem logical to have the roll on a similar set of keys to the other camera controls to make it more fluid.

PR #45, opened 2024-10-08
FIRE-34629 Add an "eye" button on the login screen allowing visible password text

SimonLsAlt:
Add an "eye" button on the login screens allowing visible password text

See https://jira.firestormviewer.org/browse/FIRE-34629 for what this looks like.

This re-uses an existing "eye" button graphic from group permissions.   To be perfectly neat and clean, either this feature should have it's own button graphics, or the existing ones should be renamed to be something more generic.  That said, it works and has minimal disruption of other code and doesn't add more files and data to the download.   It didn't seem worth the effort.

I did not touch the Linden login files since I assume that would be better left alone and might do a patch for that viewer someday

### DIFF
--- a/indra/newview/fspanellogin.cpp
+++ b/indra/newview/fspanellogin.cpp
@@ -191,6 +191,7 @@ FSPanelLogin::FSPanelLogin(const LLRect &rect,
     setBackgroundOpaque(true);
 
     mPasswordModified = false;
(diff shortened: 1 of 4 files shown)


beqjanus:
I'm going to merge this, there's some tweaking to do on the alignment. I might even move it. I probably want to change the placeholder password too

PR #46, opened 2024-10-13
Fire 30873: Another Poser

AngeldarkRaymaker:
I present some new work. While some of NiranV's work *was* in here, it's almost all gone. The only things that are the same are the icon and the name in the menu.

Should you choose to review: just a reminder: C++ isn't my daily driver. So if I've put in a construction that seems weird, it's almost certainly ignorance rather than inspiration.
You'll note I've largely gone without millions of locals; I wanted it stateless as possible, and particularly prefer code behind the UI to be that way to save synchronization headaches.

So..
The gist:
- there's very little contact with the LL codebase, to minimize conflicts down the line (one cpp, 5 xmls & cmakelists)
- Posing is done with a derivative of LLMotion (fsposingmotion cpp/h), lljoint changes happen in here; it's intended this behave like any other llmotion, but rather than it lerping and slerping between a pre-defined set of frames, the user is manipulating the 'one frame' of the motion. It would be possible to 'add more frames' and make an actual animation if one wanted to, and I hope it's visible that this wouldn't be too hard to do, but, I don't machinima, so, for another day.
- there's a 'shim class' between the UI-code-behind and the llmotion derivative; this provides for fun things we want to do with gettering and settering of the rotations/positions/scales the motion offers: its purpose is to abstract all of the interactions with the motion. Hopefully this is intended to keep the code-behind the UI tidy and the lljoint stuff in the llmotion derivative.
- It was mentioned the skeleton may change, one day, there's insulation against that, but of course we can't completely remain isolated and have something that makes sense to read

What's it got that's different?
- A hopefully less intimidating UI
- organized joints list, in several separate lists, clustered by area
- Undo/Redo the edits you make
- It resets the positions/scales/collision volumes to initial conditions when you stop posing them or that joint
- 2D slider control (trackpad) for rotation changes (try it, I think it's quite nice, it's not the azi-elev sun/moon control, that thing is mental! it got dumped for my own UI control very early on; one may find the trackpad could neatly replace other camera controls, because we need more camera controls (ha ha Beq!) It largely works like ctrl-mousing the camera does now)
- simplified language: up/down/left/right/roll instead of X/Y/Z
- friendlier joint-names, pulled from XML (so can regionalize perhaps?)
- Collision volumes that actually work
- 'the opposite of mirror' changes: 'sympathetic': When you move one arm right, the other can go left (mirror) or right as well (sympathetic)
- Pose-directory management (opens explorer/finder window on your pose dir)
- option to save BVH as well when saving pose (though BVH is not that nice, it's something)
- not every joint/collision volume is listed: it's intended that we only provide the things that do something useful
- 39% less passive-aggressive and *completely fucking obvious* comments in the code 👶 But seriously, I put function descriptions on everything useful, used much longer function names, and tried to keep methods as short and sharp as possible. I hope this makes reviewing 4k LOC less tedious

What's it got that's the same:
- most of the joint rotation/position/scale controls one might expect from BD
- flip/mirror/stop/start/recapture/reset joint and/or motion
- load save into XML format; BD poses mostly load fine, though the finger rotations appear axially different, I didn't care enough to fix it
- the same pose dir to load/save from (meh, why not)


beqjanus:
Thankyou @AngeldarkRaymaker as discussed in-world, I like the way the UI looks now a lot more than the previous version. I'll have a look at the changes tomorrow. I might squash the commits as I'm not sure if the full commit history adds value? 
This is a lot of work, thank you for taking it on.

chanayane:
Oh now that is interesting and exciting!

AngeldarkRaymaker:
> Oh now that is interesting and exciting! Is it stable enough to try it on? Can it coexist with old poser?

Stable, I think so. I've used it for many hours at a time, on several occasions. It's ready for beta-testing & UI feedback.

Having the two of them: I've not designed them to sit side-by-side. I think the image may conflict on merge (pick either, they are the same) and the menu-items be confusing (if they conflict, you'll need to resolve such that you have both menu options). If you dealt with those (should they appear), then it will likely work: the two function differently. I'm fairly sure all the boost::bind and UI stuff is distinct. All the files are distinct. The two may fight for animation priority if you used them both at the same time (I'd not do that).

AngeldarkRaymaker:
> Thankyou @AngeldarkRaymaker as discussed in-world, I like the way the UI looks now a lot more than the previous version. I'll have a look at the changes tomorrow. I might squash the commits as I'm not sure if the full commit history adds value? This is a lot of work, thank you for taking it on.

You're welcome.
I forgot to mention, by design, I'm expecting it to work on OpenSim, since it doesn't depend on an existing asset (an animation AssetId is gotten from the server in the same way the BVH preview does). I have not tested on Open Sim though.

AngeldarkRaymaker:
Another thing I forgot to mention: the track pad UI control I added. It has the option of having 'two thumbs', that is: two moveable dots (one with left mouse, one with right mouse). A feature in search of a function.

I added this for a lark, but it's occurred to me that it could allow for direct manipulation of a quaternion (rather than going through Euler angles). I've not tried it, it could just be quackery, but the control (as expected) has issues with gimbal locking when any angle gets up about 80+ degrees... but mad-mousing two thumbs to wiggle a quat sounds like a great way to spend a Friday night.

beqjanus:
I've done some work today, cleaning up things like boost for lambdas, replacing catch... with std::exception, etc. I've been using the poser a little and found it to be hypersensitive and a little hard to control, even with the magnifier on. However, I did later find that there's possibly a setting to reduce that in the advanced settings; I haven't played with that yet.

chanayane:
I will try it in my viewer, merged your PR together with the BD poser. I had to solve a few conflicts but nothing difficult. Of course I won't use both in same time :)

chanayane:
Well this is quite nice already! It works well and I favor that UI over the BD one. I integrated it in my viewer. The trackpad is definitely a better idea than the sliders. Maybe it would be good to go slower when shift-clicking in it? Good work Angeldark ;)

chanayane:
Making it live together with the BD poser was easy, I had to rename some enums and enums members in BD code, and adjust some XML files so that they both use 100% different files.

beqjanus:
yes the level of control is quite jumpy and the damping on it seems to cause me to overshoot, so that's something I was pondering. My thoughts at the moment are that I'll get this merged after I ship the deltaFPS release as I want/need it to have a good run through the beta

chanayane:
> Would the * be for any change (rotation/position/scale?)
Yes, but maybe there is a better way of showing it...

chanayane:
I think it's not easy to roll the wheel while holding left button, to change the third axis. Maybe provide an alternative way to do that, like holding a modifier key? :)

AngeldarkRaymaker:
Currently working on a config-setting and a slider for user-control of trackpad sensitivity.

AngeldarkRaymaker:
> I think it's not easy to roll the wheel while holding left button, to change the third axis. Maybe provide an alternative way to do that, like holding a modifier key? :)

Yes, the mouse is a two-axis controller, a wheel offering a third.
The plumbing for MASK is already in handleHover... it's just what to do with it?

Could add sensitivity, though then it becomes a mildly hidden feature (arguably wheel is?).

If one presses, say, shift, what would happen?
- moving mouse only operates roll
- one of the axes changes: so you go from changing up/down/left/right to up/down/roll?
- now that I think on it, this is much like what happens when you use ctrl/alt while mouseDown to pan or zoom the view

I think that's probably the answer: do something like what the viewer does... thoughts?

chanayane:
Up, down, left, and right are the most useful axis so they should be the current axis when no modifier is held.
Then I think we could go with two modifiers.
- Shift would switch to roll axis only
- Ctrl would reduce sensivity temporarily
These modifiers can be specified in the trackpad tooltip.

Something else that would be very nice, I don't know if that's possible: being able to open/close the 4 fingers at once, only up and down axis (thumb would be apart).  Base, middle and tip bones would be moved in same time, until base and middle are both at 90° angle, then only the tip would move. That would save so much time !

AngeldarkRaymaker:
> Something else that would be very nice, I don't know if that's possible: being able to open/close the 4 fingers at once, only up and down axis (thumb would be apart). Base, middle and tip bones would be moved in same time. That would save so much time !

You can already select multiple body parts (ctrl or shift click, like other things).
ctrl-clicking all of the finger-bases, then changing rotation gets some of the way there.

But, what I have been toying with is a 'delta mode'.
Presently when you adjust multiple joints, they all get the same position/rotation/scale: all the same value...
In 'delta mode', I could change each joint from it's current position by the relative amount you move the cursor (rather than setting them all to the same). It's a super-advanced feature, but would do what you're talking about much more readily.

beqjanus:
Longer term, do we need a custom hand and even face tab that lets you activate the ~~right~~ correct bones?

AngeldarkRaymaker:
Long term: hmmm...
- presets: like some body-huds & hand-pose huds have; not perfect for different rigs. But! In preset mode, there are no llJoints, just a choice of presets for each hand. Presets dovetail into advanced usages. Commit wise we're looking at a some collection of XMLs perhaps?
- 'simple' mode: one slider for 'splay' of ALL fingers and one slider for 'grip' of ALL fingers, then 'advanced' mode which is what is there now?
- Hide the hand/face tab all-together unless one chooses advanced mode
- renaming joints? I chose base/mid/tip in a pinch... are there better names?

There is a lot of space left with that vertical tabgroup.

It's really a question of who are we pitching the UI at? What 'user proficiency'? Then do we wish to differentiate basic/advanced in this part of the floater?

A bit problem with fingers is they really suffer from 'axis drift'. One needs to add roll as one changes up/down/left/right to stop them looking derpy. That's one of the reasons I love wheel for roll: I can do 3-axis changes simultaneously. 3-axis is also useful for working around a gimbal-lock. I'm an advanced user though and am used to an hour for finger posing.

chanayane:
I like the 3 axis with mouse wheel too, but would it be possible to use it without clicking a mouse button in same time? Like as soon as the mouse cursor is hovering, we could use the wheel

chanayane:
I like the new trackpad changes :)
I think when holding Shift or Alt, or releasing them, the position on the trackpad should be updated to the current active axis.

AngeldarkRaymaker:
> I think when holding Shift or Alt, or releasing them, the position on the trackpad should be updated to the current active axis.
I thought to do this... but the math is tedious, I didn't really feel it added much when I tried.

I finished up work on the 'delta mode' I was talking about.. it's interesting, another kind of Mirror/Sympathetic.. it might make some natural looking changes.
I updated BVH saves, they are much better now. I might try saving the finger/face out to BVH, I don't know if that works on reload?

At the moment I'm thinking about adding pre-set options to the hands/face tabs... It should be quite easy as it's just loads. I may revisit loading options in general: they're very BD and unintelligible.
I also want to come up with a TOS-compliant way the poser could work with others. Not necessarily posing someone else, but more like my poser telling your poser what my pose is: so posing is collaborative. I'd pose me, you'd pose you, and we could see each other's poses in-world. It's just a thought: not something to leap in to.
I love the idea of photographers/machinimamist working together. It's quite niche what I'm thinking of, but could make some crazy fun times while genuinely collaborating on a scene together, or in a group.

AngeldarkRaymaker:
Added horizontal tabs in the Hands tab.
Added Presets tab for hands, in addition to the existing adjustments.
It looks in a subdir of your User poses dir. shown to populate presets.
It loads the same XMLs the Poser saves, even XMLs for entire poses, all it does is loads the hands, so should be end-user friendly.

A user should be able to readily make their own hand poses and get them into the hand_presets by file management, or some collection of presets could be shipped with the installer?

image

beqjanus:
Now that the 7.1.11 release is out of the way I am going to merge this in. That will give us time to get it tested and cleaned up where appropriate.

chanayane:
Oh yay this is exciting! :)

PR #47, opened 2024-10-17
Add teleport here button

makidoll:
This PR adds a button that lets you teleport using the context menu. I find this a better option than the double click teleport because some of the scripts I interact with  causes an accidental teleport from clicking too fast.

I updated the land context menu and object context menu. This doesn't show up in the pie menu.

Screenshot From 2024-10-17 05-00-10

Screenshot From 2024-10-17 05-00-19

### DIFF
--- a/indra/newview/llviewermenu.cpp
+++ b/indra/newview/llviewermenu.cpp
@@ -5796,6 +5796,23 @@ class LLLandCanSit : public view_listener_t
     }
 };
 
+// <FS:makidoll> Add teleport here button
+class LLLandTeleportHere : public view_listener_t
+{
+    bool handleEvent(const LLSD& userdata)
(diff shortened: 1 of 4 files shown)


MelodeyOwo:
Isn't this redundant given we already have the tp2cam shortcut which is arguably faster than right clicking the ground?

beqjanus:
If you are camming around, right-click "tp here" is a lot easier than tp2cam. Is this any different to double-click tp though?

beqjanus:
Hi @makidoll 
Thank you for the PR, after discussing this with the support team, we don't feel that this is a change we can merge without it adding more confusion to the menus.

I do appreciate the contribution, sorry that we can't adopt it this time.

PR #49, opened 2024-10-28
FIRE-34771: Remove BVH save option from poser

AngeldarkRaymaker:
Strip BVH in toto.
Can revert if it's ever wanted again.


beqjanus:
Yep, I've been contacted by LL on this too. So BVH goes til we do the delta thing

PR #54, opened 2024-12-03
Fire-34884: Adds BVH save, breaks gimbal lock and a few bug fixes & usability tweaks

AngeldarkRaymaker:
Two big things and a bunch of little things.

**BVH is back:** It'll only write out 'deltas', and to write out non-zero angles to BVH, they'll need to be 'base-rot = zero'.
BVH now be saves out the correct rotation and in the right order; both were incorrect in the prior implementation (angle order makes a world of difference).
Setting the base rot to zero makes BVH much more WYSIWYG when uploading. Not editing the joint leaves it 'free', so allowing it to 'play on top' of other animations. If I read right, the BVH preview deserializes BVH and reserializes it to ANIM for consumption by LLKeyframeMotion... could be a neat shortcut in there for the future?

**I've added a collection of files** to get included in the installer: a directory of XMLs which are hand poses. These poses were made by me from T-Pose. My implementation for conveying them to the user may be a little squirrely, I couldn't find a nice example of files being moved from Install to User-dir, so:

1. The poses are in the repo at indra/newview/poses/hand_presets and are to be delivered with the installer
2. At OnOpen() time, createUserPoseDirectoryIfNeeded(); file-copies the XMLs out of the install location to the right place

If this isn't how one does it, please let me know. I've not set this up to work with Visual Studio, viewer_manifest.py just does it for is_packaging_viewer. It shouldn't explode in VS without them.

I added the XML to CMAKELISTS for completeness, but not sure it's needed, let me know.
There's more remarks in the JIRA https://jira.firestormviewer.org/browse/FIRE-34884


indra/newview/fsfloaterposer.cpp
```diff
@@ -1395,18 +1444,33 @@ std::vector<FSPoserAnimator::FSPoserJoint*> FSFloaterPoser::getUiSelectedPoserJo
...
+{
+    if (jointName.empty())
+        return ABSOLUTE_ROT;
+
+    bool hasRotationStylePreferenceParameter = hasString(XML_JOINT_DELTAROT_STRING_PREFIX + jointName);
+    if (!hasRotationStylePreferenceParameter)
+        return ABSOLUTE_ROT;
+ 
+    std::string paramValue = getString(XML_JOINT_DELTAROT_STRING_PREFIX + jointName);
+    if (strstr(paramValue.c_str(), "true"))
```
Ansariel, line 1457: `if (paramValue == "true")`

indra/newview/fsfloaterposer.h
```diff
@@ -176,7 +186,7 @@ class FSFloaterPoser : public LLFloater
     /// There may be +/- PI difference two axes, because harmonics.
     /// Thus keep your UI synced with less gets.
     /// </remarks>
-    void setSelectedJointsRotation(F32 yawInRadians, F32 pitchInRadians, F32 rollInRadians);
+    void setSelectedJointsRotation(LLVector3 absoluteRot, LLVector3 deltaRot);
```
Ansariel, line 189: `void setSelectedJointsRotation(const LLVector3& absoluteRot, const LLVector3& deltaRot);`

indra/newview/fsposingmotion.cpp
```diff
@@ -266,12 +266,35 @@ bool FSPosingMotion::allStartingRotationsAreZero() const
 void FSPosingMotion::setAllRotationsToZero()
 {
     for (auto poserJoint_iter = mJointPoses.begin(); poserJoint_iter != mJointPoses.end(); ++poserJoint_iter)
+    {
         poserJoint_iter->zeroBaseRotation();
+        poserJoint_iter->setRotationDelta(LLQuaternion::DEFAULT);
+    }
+}
+
+bool FSPosingMotion::vectorsNotQuiteEqual(LLVector3 v1, LLVector3 v2) const
```
Ansariel, line 275: `bool FSPosingMotion::vectorsNotQuiteEqual(const LLVector3& v1, const LLVector3& v2) const`

indra/newview/fsposingmotion.cpp
```diff
@@ -266,12 +266,35 @@ bool FSPosingMotion::allStartingRotationsAreZero() const
...
+    if (vectorAxesAlmostEqual(v1.mV[VX], v2.mV[VX]) &&
+        vectorAxesAlmostEqual(v1.mV[VY], v2.mV[VY]) &&
+        vectorAxesAlmostEqual(v1.mV[VZ], v2.mV[VZ]))
+        return false;
+
+    return true;
 }
 
-constexpr size_t MaximumUndoQueueLength = 20;
+bool FSPosingMotion::quatsNotQuiteEqual(LLQuaternion q1, LLQuaternion q2) const
```
Ansariel, line 285: `bool FSPosingMotion::quatsNotQuiteEqual(const LLQuaternion& q1, const LLQuaternion& q2) const`

indra/newview/skins/default/xui/en/menu_fs_poser_poses_btn.xml
```diff
@@ -3,48 +3,32 @@
...
-    </menu_item_call>
-    <menu_item_call
-     label="Load Rotations &amp; Scales"
-     layout="topleft"
-     name="rotations_scales">
-        <on_click function="Pose.Menu" parameter="rot_scale"/>
-    </menu_item_call>
-    <menu_item_call
-     label="Load Position &amp; Scales"
+     label="Load - Rotations Only"
```
Ansariel, line 22: What is the difference to `Load Selective - Rot Only`?

AngeldarkRaymaker: "Load -Rotations only" loads rotations to all joints, regardless of their on/off state. 
"Load Selective" is a new feature, loading only the joints which are currently 'off'.

I couldn't think of a better name... yeah.

beqjanus: "Load Selected (Rot)" (or Rot only) would be a more informative version. Selective just makes you got ok, selective based on what criteria, where as selected is a commonly used term for those bones that are uhm selected.

AngeldarkRaymaker: There is lots of space... I can imagine there is a colossal German work for it. 
image

beqjanus: only joints that are off? That sounds like the opposite of what I expected, what does off imply here?

AngeldarkRaymaker: 'Off' here means the Poser is not posing the limb.

The goal of the feature is to merge 2 poses: my current pose with the one I 'load selective'.
It's fairly narrow-use, but provides a way to get 'those legs' with 'this upper body' or something similar.

I need a way to specify which limbs should be loaded selectively, and it seems 'if it is off' then that expresses intent the user isn't animating those limbs with the current pose (so could do something else with them).

There's an analogue to this function and 'Off' state of a limb: if the poser is not posing a limb, then other in-world animations (AO/poseball etc) can animate the limb(s); so this kind-of fits with that idea: "if the limb is off, then something else poses it".

kiss-spicoli:
Thanks for this!  Tested a rebuilt pose in Phoenix-FirestormOS-AyaneStorm_AVX2-7-1-12-77512.  Saves and reloads properly, upload and play worked, upload installed and positioned in furniture and played looks good!  

Screenshot 2025-01-20 134517

PR #59, opened 2024-12-31
[Fire-35019] Add LLHUDNameTag background to floating text and hover highlights

minerjr:
This pull is to add support for in world floating text to have a background based on the LLHUBNameTag object. This also adds support for when a prim which has a floating text is hovered over. The text becomes opaque and changes its render order to appear over all other floating text, as well as the background pulled forward. This also disables the floating text fade on the highlighted text so that very far away objects can be checked.

This is designed support better contrast the text from other floating text that may be near. I play DFS with my wife and there tends to be a lot of floating text. With my color blindness I had a hard time reading some of the darker text as well as seeing what item I am hovering over.

This is an example of what the our sim looks like currently with no enhancement. 
Note the large amount of text on the objects to the left of my avatar.
Also note that far away objects, the text has faded away.
FIRE-35019-01

Now with the hover over prim to highlight floating text check enabled, very far away text will appear only on objects highlighted
FIRE-35019-02

Now once the Show floating text background is changed to "Only Highlighted Prim", a white or black transparent background will appear on the highlighted prim. You can now see the status of the prim and not have to re-position your camera or move your avatar.
FIRE-35019-03

You can move the cursor to the next prim and see its status.
FIRE-35019-04

The user can also set the Show floating text background to "All Prims", which will have the background appear on all floating text.
FIRE-35019-05

You can still hover over a prim and its text and background will move to the front and also become fully opaque to be able to read.
FIRE-35019-06

In the preferences under Colors there is a new Floating Text tab, which has a slider to changing the opacity of the floating text. The default is 0.75
FIRE-35019-07

But the user can turn it down if the default is too much. Example, setting to opacity to 40%.
FIRE-35019-08

Overall changes listed below in this patch.

Added background image based upon the LLHUBNameTag background to LLHUBText.
Added flag to highlight LLHUBText attached to prims that is set when the user hovers the cursor over.
Added new value to track if user wants backgrounds on all floating text, only the one that is currently highlighted or none.
Added new Preferences->User Interface->3D World options to disable the Hover over prim to higlight floating text and Show floating text background options. Defaulted to off.
Added new Floating Text tab to the Preferences->Colors panel to control the opacity of the LLHUBText background, independent of the LLHUBNameTag value
Added three new persist values into the setting.xml file, FSHudTextShowBackground, FSHudTextBackgroundOpacity and FSHudTextUseHoverHighlight

Fixed up code commit to better follow the standards for Firestorm Viewer as well as re-factor the code to be cleaner and removed some unnecessary code. Thanks to @beqjanus for the initial feedback.

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -25106,6 +25106,41 @@ Change of this parameter will affect the layout of buttons in notification toast
       <key>SanityComment</key>
       <string>This value needs to be greater than 0 for a fading effect.</string>
     </map>
+    <!-- <FS:minerjr> [FIRE-35019] Add LLHUBNameTag background to floating text and hover highlights -->
+    <key>FSHudTextShowBackground</key>
+    <map>
+        <key>Comment</key>
+        <string>Displays a black/white background behind the prim floating text to make the text more readable</string>
+        <key>Persist</key>
+        <integer>1</integer>
+        <key>Type</key>
+        <string>S32</string>
+        <key>Value</key>
+        <integer>0</integer>
+    </map>
+    <key>FSHudTextBackgroundOpacity</key>
+    <map>
+        <key>Comment</key>
+        <string>Opacity of floating text background (0.0 = completely transparent, 1.0 = completely opaque)</string>
+        <key>Persist</key>
+        <integer>1</integer>
+        <key>Type</key>
+        <string>F32</string>
+        <key>Value</key>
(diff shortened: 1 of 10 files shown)


minerjr:
I updated the title of the Pull to reflect the fact that I fixed the typo of LLHUBText and LLHUBNameTag to LLHUDText and LLHUDNameTag.
Sometimes you cannot see the forest from the trees. :)

PR #60, opened 2024-12-31
[FIRE-35024] - Store always run flag on exit and restore when logging in

minerjr:
Added new optional flag to restore the Always Run flag on login based on the value at exit. This is based on the existing Always Fly flag. 

Also added new checkbox to use the flag on login or not (Default) in the Preferences->Move & View->Movement panel labeled "Restore Always Run on login" and store the value in as an additional flag in the settings.xml file.

Showing the new checkbox option. Now when you have Always Run enabled at logout, when you login again, it is set.
FIRE-35024-Flag Enabled

Change Summary:

Added two new persist values to settings.xml, FSRestoreAlwaysRunAtExit and FSAlwaysRunAtExit. 
Added new checkbox to panel_preferences_move.xml FSRestoreAlwaysRunAtExit, used to enable/disable the use of the AlwaysRunAtExit setting. 
Updated llappviewer to save the current always run value on disconnectviewer and to default to false on NonInteractive init.
Updated llagent to restore the always run flag only if the FSRestoreAlwaysRunAtExit flag is set by the user.

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -5612,6 +5612,34 @@
       <key>Backup</key>
       <integer>0</integer>
     </map>
+    <!-- <FS:minerjr> [FIRE-35024] Store always run flag on exit and restore when logging in -->
+    <key>FSRestoreAlwaysRunAtExit</key>
+    <map>
+        <key>Comment</key>
+        <string>Should we restore the always run flag on login</string>
+        <key>Persist</key>
+        <integer>1</integer>
+        <key>Type</key>
+        <string>Boolean</string>
+        <key>Value</key>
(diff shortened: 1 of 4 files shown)


beqjanus:
The support team echoed my hesitance over this, so for now, I am going to decline this feature.
The concerns are:-
1) We already have over complex UI and settings and adding additional settings for what is only a single keypress in any case does not seem worth while.
2) We have no record of anyone requesting this, which supports point 1
3) People will set always run persistence and then forget how to turn it off and turn up in support whining

In addition to being a single key enablement, you can also [set AlwaysRun as part of your startup ](https://wiki.secondlife.com/wiki/Viewer_parameters) using `--set AlwaysRun true`

PR #63, opened 2025-01-02
[FIRE-35042] Inventory - Only Coalesced Filter - More accessible

minerjr:
**Issue:**

In the Main Inventory, there is a optional filter called "Only Coalesced" that only appears on the Gear button menu and does not describe what it does and is very hard to find.
FIRE-Only-Coalesecd_old_ui_02

FIRE-Only-Coalesecd_old_ui_03

Most user's know this as Clumps, and are created by entering Edit mode and selecting 2 or more prims and right clicking and in the Pie Menu selecting the option "Take: Combined". One issue is the UI filter is called "Only Coalesced", which is confusing to the user and has not tool_tip to explain what it does.
FIRE-Only-Coalesecd_old_ui_07

And it is not obvious to the user that the flag is enabled and what it did. The filter is toggled in the below screen shot.
FIRE-Only-Coalesecd_old_ui_04

If they expand a folder that contains Objects that are Combined, then they are visible.
FIRE-Only-Coalesecd_old_ui_05

Also, on the Filter dropdown, the filter "Only Coalesced" is not listed.
FIRE-Only-Coalesecd_old_ui_06

**Proposed Change:**

First rename the visible references from "Only Coalesced" to "Only Combined (Clumps)". Which makes reference to what UI choice the user used to create the combined objects. Also make reference to the nickname for the objects that users are familiar with.
Update the main inventory window to have a new button at the bottom with the Coalesced icon (Inv_Object_Multi) which can toggle on the filter, as well as visually indicate the filter is enabled.
This button also has a tool tip: "Only Combined (Clumps) - Enables filter that displays only Combined Objects (Clumps) which are made up of 2 or more prims"
Also, on the Finder window, there is now Combined (Clumps) filter checkbox with the icon of what the Inv_Object_Multi looks like.
And finally, the Filter dropdown list now has a separate option for "Only Combined (Clumps)".
FIRE-Only-Coalesecd_new_ui_01

When the user selects the Filter: "Only Combined (Clumps)", the Only Coalesced filter is enabled and the filter option "Objects" is selected. Please note that the new button is toggled on and so it the Finder checkbox.
FIRE-Only-Coalesecd_new_ui_02

The only folders that contain the Combined objects are listed and when the user expands them, they can view the objects.
FIRE-Only-Coalesecd_new_ui_03

If the user clicks either the Only Combined (Clumps) lower button or Finder checkbox, then the filter is disabled.
FIRE-Only-Coalesecd_new_ui_04

Only Combined (Clumps) disabled.
FIRE-Only-Coalesecd_new_ui_05

If the user picks Reset from the UI they are able to reset the Only Combined (Clumps) flag.
FIRE-Only-Coalesecd_new_ui_06

The user is able to select the filter, without the need to use the top Filter dropdown.
FIRE-Only-Coalesecd_new_ui_07

The lower Only Combined (Clumps) button stays highlight to let the user know the filter is enabled.
FIRE-Only-Coalesecd_new_ui_08

The original Only Combined (Clumps) still works as before, but is synced with the other UI features.
FIRE-Only-Coalesecd_new_ui_09

Once selected, the Finder checkbox and the lower button set to their toggled state.
FIRE-Only-Coalesecd_new_ui_10

**Change List:**

Added new button to panel_main_inventory.xml called only_coalesced_inv_btn and panel that can toggle on the Only Coalesced filter and added a tool_tip as well to explain the feature.
"Only Combined (Clumps) - Enables filter that displays only Combined Objects (Clumps) which are made up of 2 or more prims"

Added new combo box item filter_type_coalesced to Filter to allow setting the Only Coalesced flag and switching the filter to Objects as only objects can be coalesced.
Added new checkbox with icon to the float_inventory_view_finder.xml that is linked to the Only Coalesced flag and can modify it.

Changed label from "Only Coalesced" to "Only Combined (Clumps)" to make it more descriptive and match the UI verbiage that creates the Coalesced objects.
Updated UI xml for all English skin variants.

Added new texture coalesced_18.png for the new button. Gray Scaled version of the Inv_Object_Multi object - inv_item_object_multi_.tga with a size of 18x18 pixels.


beqjanus:
Thanks for this, I like the general idea, and I agree that consistency is important. I'll speak to the support team about the terminology we want to use. It needs to be consistent and it needs to translate well and then I'll take a closer look at importing the changes and come back to you with any extra feedback (if needed)

minerjr:
Cool, thanks for the look at the pull.

beqjanus:
I spoke with the team and the general consensus is aligned to my thoughts, we should not be changing the accepted language (where coalesced is the term that has always been used) moreover combined does not really improve things, having spoke to our non-native English teams coalesced works in translation as much as combined does and clumps does not translate well at all. 

The other issue that irks me is the icon placement. I do quite like the presence of the icon, it helps underline the meaning, but placing it over on the left there makes the alignments wrong :-) is it possible to move that icon to be in the label I wonder? I think that might be a struggle, which is probably why you placed it in the awkward place it is in now. I think if it can be moved to the end 'Coalesced ([icon]' perhaps, then it works, if not remove the icon for now and we can perhaps explore how to get that back in later.

If you would like to update the language to be coalesced again then I think we can look at merging this. 

Please refine this PR if you can. I think you closed the previous one and recreated, but if possible update this PR as it keeps the history intact.

Nice work, I am liking the ideas, even if not all of them appeal to me having new ideas and contributions is great to see.

Ansariel:
I personally do not like the icon for coalesced items into the main inventory "status" bar. Filtering for those items is rather an edge case and I don't think it justifies its prominent placement there and should just remain in the search/filter menu together with the other options.

minerjr:
Hi @beqjanus and @Ansariel, I have revised my pull request with the feedback I received from both of you. I have reverted the filter UI text back to the original.
@Ansariel your feedback was spot on when it came to the Only Coalesced filter on the bottom button panel and I did a second iteration of the concept.

I changed the single Only Coalesced filter to a new Show Filters button.
FIRE-35042-revisoin01

When you click on it, it opens the Finder/Filters side menu.
FIRE-35042-revisoin02

Here you can see I swapped the Location of the Only Coalesced check box and Icon Left<->Right.
There is also a new combo box added that adds the missing Show Links filter from the menu.
FIRE-35042-revisoin03

When you click on ANY filter, the Show Filters button on the main menu now becomes toggled and stays highlighted.
FIRE-35042-revisoin04

The Finder window is synced with the other existing menus and windows and the Show Filters button stays highlighted.
FIRE-35042-revisoin05

The new Show Links combo box includes all three states for the Show Links filter and follows the order in the gear menu.
FIRE-35042-revisoin06

Once selected from the combo box, the filter is applied (I have no linked items, so my inventory is showing empty)
FIRE-35042-revisoin07

If you close your inventory now and open it again, the user will see that they have no items. 
Normally this would be confusing, but now we have the still highlighted filter button which pressed will show the filter menu again.
FIRE-35042-revisoin08

If the user turns on a filter for example Modify filter, the filter icon will turn on.
FIRE-35042-revisoin09

If the user touches the Show Filter button, it will close the Finder window, but stay highlighted.
FIRE-35042-revisoin10

Using the "Reset" button either on the Finder menu or the Gear menu will clear the filter and reveal any previous hidden items.
FIRE-35042-revisoin11

I looked at adding a Preferences to hide the button, but ran into XUI issues with removed children or visibility=false children not refreshing the UI dynamically and there was only a hole in the bottom panel where the hidden button was. So I figured I make it a universal button and that may be better for everyone.

beqjanus:
I like this a lot, it feels more complete. I'll look at merging it in later.

PR #71, opened 2025-01-22
Allows to limit search to folders only in inventory

chanayane:
This lets you limit your search to folders only. Very useful when you have a large inventory and you need to find a specific folder without the search getting drown with folder items.

It does not change the default behavior, if you uncheck "Include items" it will limit the search to folders only and show only folders.

Normal behavior, "Include items" checked
FirestormOS-AyaneStorm_VM223Ybhq7

"Include items" unchecked
FirestormOS-AyaneStorm_hpCm82bHoh

### DIFF
--- a/indra/newview/llinventoryfilter.cpp
+++ b/indra/newview/llinventoryfilter.cpp
@@ -702,17 +702,29 @@ bool LLInventoryFilter::checkAgainstSearchVisibility(const LLFolderViewModelItem
     if (!object)
         return true;
 
-    const bool is_link = object->getIsLinkType();
-    if (is_link && ((mFilterOps.mSearchVisibility & VISIBILITY_LINKS) == 0))
+// <AS:chanayane> Short circuit search optimisation
+    //const bool is_link = object->getIsLinkType();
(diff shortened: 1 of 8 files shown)


chanayane:
I also optimized the `LLInventoryFilter::checkAgainstSearchVisibility` method to short-circuit so as to avoid unnecessary loops.

beqjanus:
Silly question maybe, but would a filter of "folders" (as a dummy asset type) be more logical?
In the current setup I could filter to "Clothing" and then only match on folder names. I'm not sure if that makes sense to a user or not?

chanayane:
> Silly question maybe, but would a filter of "folders" (as a dummy asset type) be more logical? In the current setup I could filter to "Clothing" and then only match on folder names. I'm not sure if that makes sense to a user or not?

Yes, I will rework this.

chanayane:
I reworked this and used the filter combobox instead.

FirestormOS-AyaneStorm_wnxf9Y1Wuo
FirestormOS-AyaneStorm_zGFYktawNX

PR #72, opened 2025-01-23
FIRE-35083 - Floater_stats.xml has incorrect stat_view setting for materials

minerjr:
The issue was caused by the incorrect setting being applied to the material stat_view and child stat_bar. Added the correct named setting for the open/close state of the material stat_view and updated the float_stats.xml to have the correct settings applied. This prevents the warning and also allows the materials stat_view to have its own/close state saved correctly.

Stats menu option
float_stats_menu

Just opening the Statistics Panel in debug caused the following bug splat:
float_stats

In VS in release you can see there is a warning being generated and returns false
float_stats_vs

Which causes the stat_view for materials to always be close upon opening the window
float_stats_window

Corrected the issue by correcting the float_stats.xml to use a proper boolean settings.xml OpenDebugStatMaterials flag for the state of the window, as well as properly use the DebugStatModeMaterials to the settings of the child stat_bar.

Now restores the material view open upon opening the Statistics floater window now.
float_stats_vs_fixed

float_stats_window_fixed

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -9242,6 +9242,20 @@ Change of this parameter will affect the layout of buttons in notification toast
       <key>Value</key>
       <integer>1</integer>
     </map>
+    <!-- <FS:minerjr> [FIRE-35083] Floater_stats.xml has in correct stat_view setting for materials -->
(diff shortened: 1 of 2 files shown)


minerjr:
Going to check in directly to the master codebase.

PR #73, opened 2025-01-28
Double click on animation in AO

chanayane:
This PR adds the ability to double-click on an animation in the AO panel, and it will play it directly. This is more convenient than clicking many times on the Previous, Next or Reload buttons.

If the AO set is not enabled while double-clicking an animation, it will get enabled automatically.

Here's a little demo of what it does:
https://github.com/user-attachments/assets/c57d040a-ca96-4a49-b62e-7d3a3dfc49e6

### DIFF
--- a/indra/newview/aoengine.cpp
+++ b/indra/newview/aoengine.cpp
@@ -960,6 +960,112 @@ void AOEngine::cycle(eCycleMode cycleMode)
     }
 }
 
+// <AS:Chanayane> Double click on animation in AO
+void AOEngine::playAnimation(const LLUUID& animation)
+{
+    if (!mEnabled)
+    {
+        return;
+    }
+
+    if (!mCurrentSet)
+    {
+        LL_DEBUGS("AOEngine") << "cycle without set." << LL_ENDL;
+        return;
+    }
+
+    // do not cycle if we're sitting and sit-override is off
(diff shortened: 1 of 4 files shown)


indra/newview/aoengine.cpp
```diff
@@ -960,6 +960,112 @@ void AOEngine::cycle(eCycleMode cycleMode)
...
+        return;
+    }
+
+    mAnimationChangedSignal(LLUUID::null);
+
+    // Searches for the index of the animation
+    U32 idx = -1;
+    for (U32 i = 0; i < state->mAnimations.size(); i++)
+    {
+        LLUUID* id = &(state->mAnimations[i].mAssetUUID);
```
Ansariel, line 1030: Why make a pointer if you dereference it directly next line? Just do
`if (state->mAnimations[i].mAssetUUID == newAnimation)`
instead.

chanayane: I am asking myself "Why?" as well! :-P

chanayane: Fixed.

chanayane:
I also fixed a crash for when we double-click on an animation in a state different from active state (like double-clicking on a "walking" animation when we are standing).

beqjanus:
Another nice tweak. Thankyou Ayane

PR #75, opened 2025-01-29
Fire 34943: Add spinners to sliders for finer adjustments

AngeldarkRaymaker:
I know I can push straight to master, but as this is a little larger, thought I'd go this route.
It's mostly changes to UI controls, adding a spinner to every slider:
image

If you're content, I can patch it over instead of PR.

The callback for the spinners a little ugly: the neat way of binding to a gSavedSettings would make this much cleaner by taking care of the refresh complexity... but having a setting for the rotation/position/scale vectors seems odd...

I chose dirty code over dirty settings, happy to rework though, I feel sure the code would be much cleaner, it'd just have weird debug settings I guess


minerjr:
Hey, I wrote a small patch for your submission where you can use a parameter for the Poser.CommitSpinner OnCommitSpinner so you can pass in an ID. You can then use a switch statement on the OnCommitSpinner and each action will only take 1 step to get to instead of the last one taking 12 branch checks. I can submit it to the PR if you want.

PR #78, opened 2025-02-04
FIRE-35118 - remove lag event for "Delete All" on notices

SimonLsAlt:
This eliminates a lag event when pressing "Delete All" in the notification window.    The UUID and type of notice are stored in a FIFO queue instead of being directly removed.   Then a new idle() routine will delete one notice per frame.

## Firestorm Pull Request Checklist
Thank you for contributing to the Phoenix Firestorm Project.
We will endeavour to review you changes and accept/reject/request changes as soon as possible. 
Please read and follow the [Firestorm Pull Request Guidelines](https://github.com/firestormviewer/phoenix-firestorm/blob/master/FS_PR_GUIDELINES.md) to reduce the likelihood that we need to ask for "Bureaucratic" changes to make the code comply with our workflows.


indra/newview/llappviewer.cpp
beqjanus, line 1: When we have a single line modification the self terminating tag should be appended to the line it applies to
```c++
// <FS:SimonLsAlt/> FIRE-35118 - Add idle() processing to delete notifications
#include "llfloaternotificationstabbed.h"
```

should be
```c++
#include "llfloaternotificationstabbed.h" // <FS:SimonLsAlt/> FIRE-35118 - Add idle() processing to delete notifications
```

beqjanus, line 1: likewise 5670-5671

indra/newview/llfloaternotificationstabbed.cpp
beqjanus, line 1: When you have a block change it is imperative that the block is terminated by a closing tag
```c++
    // <FS:SimonLsAlt>  FIRE-35118 Linden viewer just calls onItemClose() for each item
    //  which can turn into a lag event several seconds long.   Firestorm will queue
    //  notificatsion UUIDs to close them one by one in the new idle() function below.
    mDeleteNotificationsTimer.reset();
    ```
    should have 
    ```c++
    // </FS:SimonLsAlt>
    ```
    I see what you are doing here, marking up areas that you have worked in, but the tagging should be wrapped as closely to the changes as possible. leavi ng all unchanged code "outside the tags"
    so in this instance line 374 strictly speaking needs a tag but I'd let that go I think as it wold be odd.
    however, line 377 onwards correctly comment a line and that insert a block but the line immediately before the `// onItmeClose(....)` should be an opening tag `// <FS:SimonLsAlt> blah blah`

indra/newview/llfloaternotificationstabbed.h
beqjanus, line 1: 101-103 can contract down to a single code line with the self-closing tag plus an extra blank line is fine for clarity.
172-177 great

beqjanus (commented):
All good just a few more tidy ups on the tagging. 
this will seem very nit-picky, but pulling you up now on a smallish change makes it less awkward on a larger change.

PR #82, opened 2025-02-18
[FIRE-10419] Selectively process devices based on CLSID to prevent stuttering

DaxDupont:
Hey people,

First contrib so please be patient! 
I have fixed:
[FIRE-10419] (Reported in 2013)
https://jira.firestormviewer.org/browse/FIRE-10419
and
[FIRE-32546]
https://jira.firestormviewer.org/browse/FIRE-32546 which is a dupe.

So the problem is that some devices and applications will cause windows to send out device notifications frequently, this causes a 10-15 second stutter whenever this happens, this can be every few minutes. 

So what i've done is basically register for device notifications, then filter out what we need based on CLSIDs to only init those, 
This based on https://jira.firestormviewer.org/secure/attachment/22091/joystick2.patch by Casper Warden but with a bunch of changes. For one I dropped the sensors for now as we don't need them, cleaned up the code a little and made some changes for better performances I think, but performance lottery. 

Test case for existing functionality still working:
1. Attach controller/joystick
2. Move around, if its works its detected

Test case for testing if lag is fixed requires a PC with a device that will be very 'chatty' like a Valve Index in my case, basically just idle.

I've tested both the fix for the stutter and with a controller, all works.
Not sure if we're expecting any other CLSIDs or if my optimization is overkill and we should keep everything within the main loop.

### DIFF
--- a/indra/llwindow/llwindowcallbacks.cpp
+++ b/indra/llwindow/llwindowcallbacks.cpp
@@ -184,7 +184,7 @@ bool LLWindowCallbacks::handleTimerEvent(LLWindow *window)
     return false;
 }
 
-bool LLWindowCallbacks::handleDeviceChange(LLWindow *window, bool deviceRemoved)
+bool LLWindowCallbacks::handleDeviceChange(LLWindow *window, bool deviceRemoved) // <FS:Dax/> [FIRE-10419] Added deviceRemoved bool to prevent reinitialize on disconnect.
 {
     return false;
 }
--- a/indra/llwindow/llwindowcallbacks.cpp
+++ b/indra/llwindow/llwindowcallbacks.cpp
@@ -184,7 +184,7 @@ bool LLWindowCallbacks::handleTimerEvent(LLWindow *window)
     return false;
 }
 
-bool LLWindowCallbacks::handleDeviceChange(LLWindow *window)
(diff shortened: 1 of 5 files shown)


beqjanus:
Thanks for this, I may have the perfect tester as my partner has an old machine that due to a fairly broken windows (I suspect) will hang FS long enough to force a disconnect if she plughs her phone in! I'll have a look at this later today hopefully.

beqjanus:
In the meantime, I notice that there's no change tags in the code. Can you please have a look at [the Firestorm contribution guidelines](https://github.com/FirestormViewer/phoenix-firestorm/blob/master/doc/FS_PR_GUIDELINES.md) and add the necessary comments. Feel free to ask if you have any questions or look at any of the recently accepted PRs to see how others have done this.

DaxDupont:
> Thanks for this, I may have the perfect tester as my partner has an old machine that due to a fairly broken windows (I suspect) will hang FS long enough to force a disconnect if she plughs her phone in! I'll have a look at this later today hopefully.

I have been able to replicate it on another device as well that it works but more testers the merrier

beqjanus:
Finally, tested this with my partner and confirmed this fixes her crashes/hangs which I believe also relates to FIRE-32546.
Her windows explorer still hangs upon insertion of the USB device but FS is nor immune to it.

beqjanus:
Merged and added you to the contributors list. Thanks @DaxDupont

PR #83, opened 2025-02-23
[FIRE-35185] Profile preview removing embedded URLs from 1st Life tab

chanayane:
This disables rendering links when a pick or 1st life texts are in edit mode (white background, only for own profile).
It makes the preview feature work as intended.
This actually had nothing to do with the preview feature, but it allowed to notice that old bug more easily.


chanayane:
This makes links in picks and 1st life modifiable again (until then, they couldn't be modified once saved as they were rendered).

That way it works like in the Profile tab.

PR #84, opened 2025-02-28
Add font specification Yu-Gothic for Mac and Windows, Noto Sans CJK for Linux.

logue:
https://jira.firestormviewer.org/browse/FIRE-35204


beqjanus:
Thanks for the PR, I noted on the JIRA that this is still buried in the upstream LL develop branch so we may hold this until it flows down from them and then apply this if it it still required. I have asked @PanteraPolnocy who looks after the translators to see if there are any side-effects of this in the meantime.

PanteraPolnocy:
To avoid merge conflicts, I'd rather wait for the upstream.

Ansariel:
This PR is also outdated because of the flat UI changes and would require an update.

PR #91, opened 2025-03-13
prevents having Move Lock activated but disabled when for some reason…

chanayane:
There are some situations that can lead to having the Move Lock command and menu activated, but in disabled state (so, it shows, "Avatar > Movement > Movelock is enabled" on the screen, the command is activated, and the menu item is checked... But... It is disabled so you can't turn it off).

When that happens, Move Lock is actually not enforced, so this is inconsistent.

This issue is linked to the LSL Bridge.

Two situations so far where I faced this issue:
- After a crash, sometimes nothing seems to rez or load. Probably including the LSL bridge.
- If you set UseLSLBridge to false, then detach the bridge, then set UseLSLBridge to true again.

On the next few relogs, the Move Lock button will remain disabled, still showing it is activated.

Sometimes it remains disabled after many reboots. Sometimes it fixes itself. On my last attempt, I had to delete the whole %localappdata%\Firestorm folder to solve it.

This does not fix the underlying issue linked to the correct detection of the LSL bridge, but it should at least prevent seeing the Move Lock feature as being enabled AND disabled in same time. It turns Move Lock off and will only appear as disabled.


indra/newview/llviewermenu.cpp
```diff
@@ -5141,6 +5141,12 @@ class FSSelfCheckMoveLock : public view_listener_t
         if (LLGridManager::getInstance()->isInSecondLife())
         {
             new_value = gSavedPerAccountSettings.getBOOL("UseMoveLock");
+            // <FS:Chanayane> prevents having Move Lock activated but disabled when for some reason the LSL Bridge is not worn or not ready
+            if(new_value && !enable_bridge_function()) {
```
Ansariel, line 5145: No Java coding style please 😄 
```
if (new_value && !enable_bridge_function())
{
```

chanayane: Oops, fixed :).

PR #92, opened 2025-03-20
[FIRE-34020] [OPENSIM] Copy of address bar contents does not URL encode space in region name for OpenSim only

humbletim:
This fixes [FIRE-34020](https://jira.firestormviewer.org/browse/FIRE-34020) by ensuring that `hop://` addresses are correctly URL-encoded when copied to the clipboard from the address bar, similar to `secondlife://` addresses.

The problem was that `LLURLLineEditor::copyEscapedURLToClipboard` only knew about `http://` and `secondlife://` schemes. With this change it now handles `hop://` as well.

Testing:
1. Visit any OpenSim region containing spaces (for example) in the Region Name.
2. Copy the address bar to the clipboard (using <kbd>ctrl-c</kbd> or context menu).
3. Paste the clipboard contents somewhere and should see `%20`'s have been applied.

Thanks!

cc: @aiaustin


aiaustin:
I have been testing this PR patch and it appears to work fine.

I tested across OpenSim grids (Robust grids and standalones) and using Hypergrid to travel between them.

I also tested in Second Life to check for any possible issue there and that continues to work as expected.

beqjanus:
Thanks @humbletim

PR #93, opened 2025-03-23
[FIRE-35276] Fix for client freeze in OpenSim getting picks limit

Hecklezz:
[[FIRE-35276]](https://jira.firestormviewer.org/browse/FIRE-35276)

Fix for client freeze in OpenSim getting picks limit.

Setting the max picks limit to S32_MAX for OpenSim from my limited testing appears to have 0 downside and older viewers still handle this fine, they just don't allow additional picks to be created past 10.

If you believe using S32_MAX is excessive and it should be hard capped to a more reasonable number (such as 20 like on official secondlife which is set by their servers and told to the viewer), let me know and I'll update the PR accordingly.

### DIFF
--- a/indra/newview/llagentbenefits.cpp
+++ b/indra/newview/llagentbenefits.cpp
@@ -37,6 +37,8 @@
 #include "llviewerregion.h"
 // </FS:Ansariel>
 
+constexpr S32 MAX_OPENSIM_PICKS = S32_MAX; // <FS/> [FIRE-35276] Fix for client freeze in OpenSim getting picks limit
+
 LLAgentBenefits::LLAgentBenefits():
     m_initalized(false),
     m_animated_object_limit(-1),
@@ -213,7 +215,7 @@ S32 LLAgentBenefits::getPicksLimit() const
 {
     // <FS:Ansariel> OpenSim legacy economy
     //return m_picks_limit;
-    return LLGridManager::instance().isInSecondLife() ? m_picks_limit : LLAgentPicksInfo::instance().getMaxNumberOfPicks();
+    return LLGridManager::instance().isInSecondLife() ? m_picks_limit : MAX_OPENSIM_PICKS;
     // </FS:Ansariel>
 }
 


aiaustin:
Hi Hecklezz. Recent OpenSim dev master has made a couple of commits that communicate the OpenSim max number of picks between server and viewer.

See commits on 2025-03-14 at http://opensimulator.org/viewgit/?a=shortlog&p=opensim

Hecklezz:
> As I think you know, OpenSim dev master has made a couple of commits that communicate the OpenSim max number of picks between server and viewer. Does this fix the issue with recent Firestorm 7.1.12 Beta?
> 
> See commits on 2025-03-14 at http://opensimulator.org/viewgit/?a=shortlog&p=opensim

I did not know this, however this requires every single server to be up-to-date with this version of Opensim or newer, which most grids won't be for months.
Regarding if that change fixes the issue, no it won't. Because currently, Firestorm checks if it is SecondLife and if it isn't, it doesn't read what the server has sent the viewer and instead goes down the infinite loop of calling the same function over and over inducing the freeze/crash.
In order to account for this new change in OpenSim, additional changes will need to be made so that if the client has been sent info from the server, use that, otherwise use a predefined max.

beqjanus:
I've been waiting for input from OS, to confirm how this is going to work going forward. 
I could have just thrown in a random const but I wanted to check on the intentions in OpenSim and to understand whether there is an enforced limit. The feedback from Ubit appears to be that there was no limit enforced by OS historically (I guess if people were silly enough to add more than 10 they just never got to see them). 
@Hecklezz I'll merge this asis, (thanks) and then rework it to respect the simulator features override when present.

Hecklezz:
> The feedback from Ubit appears to be that there was no limit enforced by OS historically (I guess if people were silly enough to add more than 10 they just never got to see them). 

Oh, before if you had more the 10 picks, they would work and appear just fine, the viewer just wouldn't allow you to create any more if you had >= 10. But if you forced yourself to have more than 10 (such as with a modified viewer), they would still continue to appear and be usable without a modified viewer.

PR #94, opened 2025-03-23
[FIRE-35288] Obtain MaxProfilePicks from simulator features

Hecklezz:
[[FIRE-35288]](https://jira.firestormviewer.org/browse/FIRE-35288)

OpenSimulator very recently implemented the Simulator Feature "MaxProfilePicks", in order to determine the maximum profile picks a person can have.

The above PR implements this, and if the feature is not provided (the server is on an older version), it falls back to the new default value of 20 (previously S32_MAX but as explained by Ubit, this is not ideal for a number of reasons).

I didn't add a comment for this specific Jira issue, since it is already inside the // <FS:Ansariel> OpenSim legacy economy block, and also other Simulator Features that have been added in the past such as MaxAgentAttachments, is inside this same block.
If you want me to tag it anyway, please let me know and I can do it anyway. (Please tell me how you would want it though, because doubling up on the FS comment tag seems a bit redundant.

### DIFF
--- a/indra/newview/llagentbenefits.cpp
+++ b/indra/newview/llagentbenefits.cpp
@@ -37,8 +37,6 @@
 #include "llviewerregion.h"
 // </FS:Ansariel>
 
-constexpr S32 MAX_OPENSIM_PICKS = S32_MAX; // <FS/> [FIRE-35276] Fix for client freeze in OpenSim getting picks limit
-
 LLAgentBenefits::LLAgentBenefits():
     m_initalized(false),
     m_animated_object_limit(-1),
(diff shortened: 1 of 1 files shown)


beqjanus:
Thanks for this but I made the change last night. Looks like our changes overlapped. 
I think our solutions are pretty much identical, so I'm going to close this one.

beqjanus:
BTW - Your observation on the comment tags was spot on. We don't typically tag inside an existing tag, though I have been known to add some extra context in a comment if the change is somehow independent of the existing block.

Hecklezz:
> Thanks for this but I made the change last night. Looks like our changes overlapped. I think our solutions are pretty much identical, so I'm going to close this one.

No problem!
I noticed you put misspell my alias and put "Haklezz" in the comment in your code instead of "Hecklezz" 😜

PR #95, opened 2025-03-25
[FIRE-35292] Fix Exporting Texture getting downscaled and compression

Hecklezz:
[FIRE-35292](https://jira.firestormviewer.org/browse/FIRE-35292)
Fixes the issue described in the Jira ticket where exporting textures as PNG or TGA were getting downscaled and compressed when they weren't supposed to be. This is achieved by simply setting the discard level back to 0 before it decodes the image and saves it to disk.

_Fixing the Uploading of J2C textures is not implemented yet. This is an issue that impacts secondlife's viewer too, and as such, I have opened an issue with their version and they have already taken it on as a bug and slated for milestone 2025.04. If LL chooses not to fix it themselves, or their fix needs changes for OpenSim, I will open another PR when necessary_

(Also corrects a minor error in a comment from a previous commit)

### DIFF
--- a/indra/newview/daeexport.cpp
+++ b/indra/newview/daeexport.cpp
@@ -505,7 +505,15 @@ void ColladaExportFloater::CacheReadResponder::saveTexturesWorker(void* data)
             std::string name = gDirUtilp->getDirName(me->mFilename);
             name += gDirUtilp->getDirDelimiter() + me->mTexturesToSave[id];
             CacheReadResponder* responder = new CacheReadResponder(id, img, name, img_type);
-            LLAppViewer::getTextureCache()->readFromCache(id, 0, 999999, responder);
+            // <FS:minerjr> [FIRE-35292] Fix for textures getting downscaled and compressed           
+            //LLAppViewer::getTextureCache()->readFromCache(id, 0, 999999, responder);
+            // The above line hard coded the size of data to read from the cached version of the texture as 999999,
+            // where now we will calcuate the correct value based upon the texture's full width, height and # of components (3=RGB, 4=RGBA) and
+            // the discard level (0)). There is a choice to change the rate, but we seem to use the value of 1/8 compression level
+            S32 texture_size = LLImageJ2C::calcDataSizeJ2C(imagep->getFullWidth(), imagep->getFullHeight(), imagep->getComponents(), 0);// , F32 rate) rate = const F32 DEFAULT_COMPRESSION_RATE = 1.f/8.f;
+            // Use calculated texture_size (from LLTextureFetch::createRequest see "else if (w*h*c > 0)" statement for more info)
+            LLAppViewer::getTextureCache()->readFromCache(id, 0, texture_size, responder); 
+            // </FS:minerjr> [FIRE-35292]
             me->mTexturesToSave.erase(id);
             me->updateTitleProgress();
             me->mTimer.reset();
--- a/indra/newview/daeexport.cpp
+++ b/indra/newview/daeexport.cpp
@@ -505,14 +505,14 @@ void ColladaExportFloater::CacheReadResponder::saveTexturesWorker(void* data)
             std::string name = gDirUtilp->getDirName(me->mFilename);
             name += gDirUtilp->getDirDelimiter() + me->mTexturesToSave[id];
             CacheReadResponder* responder = new CacheReadResponder(id, img, name, img_type);
-            // <FS:minerjr> [FIRE-35292] Fix for textures getting downscaled and compressed           
+            // <FS:minerjr> [FIRE-35292] Fix for textures getting downscaled and compressed
             //LLAppViewer::getTextureCache()->readFromCache(id, 0, 999999, responder);
             // The above line hard coded the size of data to read from the cached version of the texture as 999999,
             // where now we will calcuate the correct value based upon the texture's full width, height and # of components (3=RGB, 4=RGBA) and
             // the discard level (0)). There is a choice to change the rate, but we seem to use the value of 1/8 compression level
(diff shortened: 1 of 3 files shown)


beqjanus:
I've added minerjr as a reviewer because he's been doing a lot of work on the texture scaling/not scaling issues.

minerjr:
I was able to confirm the fix worked, but I also believe I found a fix for the save J2C part. The issue was when the texture was read from the cache, it was reading a hard coded amount of data, 999,999 instead of using the LLImageJ2C::calcDataSizeJ2C to calculate the amount of data to read from the cache. @Hecklezz  Can you confirm that this works for you?

Hecklezz:
I'm not sure what you mean by the "save J2C part", since the issue with J2C's was with uploading them, not exporting them. Exporting J2Cs from my own testing saved them correctly in their original resolution since they don't need to be decoded as you can see the function ColladaExportFloater::CacheReadResponder::completed-
// If we are saving jpeg2000, no need to do anything, just write to disk
if (mImageType == DAEExportUtil::ft_j2c)

Hecklezz:
Fyi, I put up an issue regarding uploading J2C's on LL's github as well since it impacts their official viewer too-
https://github.com/secondlife/viewer/issues/3809

So the fix for uploading J2C's should be fixed on LL's end and merged into firestorm, instead of the other way around for consistency sake.

minerjr:
So are creating an original j2c in Gimp at 2K and uploading that correct? What export options are you using to create it? Or are you importing the PNG into the viewer, then saving the J2C from the the caldara export. Then re-uploading the saved J2C to test out the J2C import functionality?

What I changed was when the caldara does to export. Is it iterator over the object and gets the list of textures assigned to the prim. When it does, it reads in the data from the cache and uses the read in texture data to then then then export the data. So what may have been happening is when export the J2C, it would only have 999,999 data, where as a correct data should be closer to 2048x2048x3 / 8 + for each level, etc.

Hecklezz:
You still aren't understand haha.
I never had any issues exporting textures as J2C, ALSO, even if you specified 1 as the size instead of 999,999, it would still successfully export J2C's at full resolution.

Hecklezz:
While we are at it, we should probably add your change to FSFloaterObjectExport::fetchTextureFromCache, since that also hardcodes 999,999.

minerjr:
Oh NP, I was just trying to get my head wrapped around the issue and order of operations. I have been dealing with KDU library for J2C a little bit, (Crashes mostly) and got to know a good bit about the Fetch and Texture system that I have been messing with for the past 3 months. :) The other thing we may able to go to the actual cache header and skip the calculation and just read in the header's amount of data stored to the UUID and use that, so we know exactly the J2C is the same size as what is in the cache. Yeah updating the other export would be good option as well. Another option for the export is to generate a LLRawImage from the image->readBackRaw() which will read from OpenGL the actual texture as is, and then use KDU to a new compression of the data and save that out for export. But if it's stored in cache you can just use the cache like current.

Hecklezz:
Alright, sorry for being a bit all over the place, 3am here and got a lot of other stuff on my plate 😅
If you are happy with the current changes, I can open this up to be reviewed and merged.

Since fixing the uploading of J2C's issue I think should rely on LindenLab at the moment with my open issue with their viewer.
Unless you want to still investigate that, and then we can propose the fix to LL, and if they don't want it, then we put it into firestorm.

Hecklezz:
> and then use KDU to a new compression of the data and save that out for export. But if it's stored in cache you can just use the cache like current.
KDU is not an option for OpenSim users, only OpenJPEG, so an OpenJPEG solution would still need to be implemented, even if a KDU one is implemented.

minerjr:
Oh no worries, I get the same at 3 AM. :) OK so your using OpenJPEG, cool. Did you try exporting with my change to the cache read code and save out the J2C from an existing texture and then re-upload the J2C and see if it was the correct resolution? The other thing is it could be the formatting of the data from OpenJPEG, may have an issue on the export with a flag or something like the compression rate is too high or too low which may be handled differently when KDU. I will test it out and do an AB test between KDU and OpenJPEG and see what the exports are from the same original texture. Get some sleep and really good work so far. I know your making a lot of people happy who use OpenSim. :)

Hecklezz:
> Did you try exporting with my change to the cache read code and save out the J2C from an existing texture and then re-upload the J2C and see if it was the correct resolution?
Uploading the J2C texture back is the other half of the problem though and why I have an open issue with the secondlife viewer xD
However, checking it on disk, yes it works perfectly fine with your change but the resolution and data of it is no different (as expected), but that is just because the texture I am uploading wasn't exceeding the 999,999 limit that was previously there.

Hecklezz:
I have modified this PR to only be for fixing exporting and not important, since we will wait for LL to see what they will or won't do first, and if it needs OpenSim changes for OpenJPEG instead of KDU. 

It is now ready for review and merge.

minerjr:
I can confirm that the export the J2K exports the same source raw data for KDU as OpenJpeg2000. 
[J2CExportTests.zip](https://github.com/user-attachments/files/19495418/J2CExportTests.zip)

Both exported a 28 KB file with the Caldara save option with each library compiled.

But I am good with the fix as it is.

beqjanus:
> I have modified this PR to only be for fixing exporting and not uploading, since we will wait for LL to see what they will or won't do first, and if it needs OpenSim changes for OpenJPEG instead of KDU.
> 
> It is now ready for review and merge.

A point of clarification, there is nothing about OpenSim that requires OpenJPEG. All the official builds of FS for OpenSim are using KDU. OpenJPEG is used when compiling for SL or OpenSim when the compiler does not have an appropriate kakadu license. 

This is different to, havok, for example, where the licensing is tied to LL and SecondLife and thus Havok cannot be used in OpenSim builds. 

None of this changes the nature of the PR, but I thought I'd just clarify matters.

Hecklezz:
> All the official builds of FS for OpenSim are using KDU. OpenJPEG is used when compiling for SL or OpenSim when the compiler does not have an appropriate kakadu license.

Oh I didn't realise this, thank you for clarifying. I expected it to have been similar to the Havok situation but I suppose I was wrong.

PR #97, opened 2025-03-27
[FIRE-35147] OpenSim MiniMap rendering fixes

Hecklezz:
[FIRE-35147](https://jira.firestormviewer.org/browse/FIRE-35147)

Fixes two issues with the MiniMap when playing in OpenSim-
1. If a region is greater than 256x256, it now correctly renders the entire region on the MiniMap instead of only only rendering the first 256x256 potion.
2. When you change to a new grid, it now correctly renders on the MiniMap the regions you spawned in, instead of it being incorrectly rendered until you left and came back.

### DIFF
--- a/indra/newview/llnetmap.cpp
+++ b/indra/newview/llnetmap.cpp
@@ -477,32 +477,27 @@ void LLNetMap::draw()
                         if (pRegionImage.isNull())
                             continue;
 
-                        if (!pRegionImage->hasGLTexture())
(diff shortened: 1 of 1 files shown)


beqjanus:
Gave this a quick test and it seems to work well. Merging it for further testing feedback in the next beta.

aiaustin:
Tested before and after the fix was incorporated. This works great now on a test 2x2 varregion.

PR #99, opened 2025-04-03
[FIRE-35327] Fixes Antialising type setting under preferences

Hecklezz:
[FIRE-35327](https://jira.firestormviewer.org/browse/FIRE-35327)

Fixes the Antialiasing type under preferences from being set to the wrong value and not saving correctly.
(It was incorrectly tied to the Antialiasing sampling setting)


beqjanus:
Good catch

PR #101, opened 2025-04-03
[FIRE-35323] Fix poser values getting stuck after using visual manipulators

Hecklezz:
[FIRE-35323](https://jira.firestormviewer.org/browse/FIRE-35323)

If you had set public values (rotation, scale and position) in the Avatar and Animesh Poser, and then proceed to use the visual manipulators, it would result in the public values being set back to 0 and their results being written into the base values of the joints. This resulted in your avatar getting stuck with these changes after closing the poser.

This PR works around this by setting back the public values and correcting the base values to not have the public values written into them.
This may not be the "proper" solution, but it's a working workaround.

PS: I noticed the Poser Animator calls the same function I implemented the workaround in, so this PR needs to be properly tested to ensure there is no adverse side effects to the fix.

### DIFF
--- a/indra/newview/fsjointpose.h
+++ b/indra/newview/fsjointpose.h
@@ -215,6 +215,21 @@ class FSJointPose
             joint->setScale(mBaseScale);
         }
 
+        void restorePublicState(FSJointState previousState)
+        {
+            const LLVector3& previousPosition = previousState.mPosition;
+            const LLVector3& previousScale = previousState.mScale;
+            const LLQuaternion& previousRotation = previousState.mRotation;
+
+            mBasePosition -= previousPosition;
+            mPosition = previousPosition;
+            mBaseScale -= previousScale;
+            mScale = previousScale;
+            mBaseRotation = ~previousRotation * mBaseRotation;
+            mBaseRotation.normalize();
+            mRotation = previousRotation;
+        }
+
       private:
         FSJointState(FSJointState* state)
         {
--- a/indra/newview/fsjointpose.cpp
+++ b/indra/newview/fsjointpose.cpp
@@ -150,15 +150,20 @@ void FSJointPose::recaptureJointAsDelta()
     {
         return;
     }
-    
+
     LLJoint* joint = mJointState->getJoint();
     if (!joint)
     {
         return;
     }
-    
+
     addStateToUndo(mCurrentState);
-    mCurrentState = FSJointState(joint);
+    FSJointState newState = FSJointState(joint);
+
+    // Fixes issue when using the visual manipulators after setting Move/Scale values
+    newState.restorePublicState(mCurrentState);
+
+    mCurrentState = newState;
 }
 
 void FSJointPose::swapRotationWith(FSJointPose* oppositeJoint)


beqjanus:
@AngeldarkRaymaker you'll need to take a look at whether this overlaps with your work in progress updates, but also that it does not open up any theft options that we need to be aware of.

indra/newview/fsjointpose.h
```diff
@@ -215,6 +215,21 @@ class FSJointPose
...
+        {
+            const LLVector3& previousPosition = previousState.mPosition;
+            const LLVector3& previousScale = previousState.mScale;
+            const LLQuaternion& previousRotation = previousState.mRotation;
+
+            mBasePosition -= previousPosition;
+            mPosition = previousPosition;
+            mBaseScale -= previousScale;
+            mScale = previousScale;
+            mBaseRotation = ~previousRotation * mBaseRotation;
```
AngeldarkRaymaker, line 228: The 'base' rotation is intended to represent the LLJoint rotation at the time we start posing.
It gets changed when a user 'recaptures' a joint, or when a user 'zeros' it, but otherwise should be largely constant.

This change looks like it would alter base, which we'd like to avoid.
This change would solve one problem, but I think create another.
We want to keep the base constant for this scenario:

1. User sits on a chair (or strikes a pose somehow)
2. User modifies the pose with poser
3. User saves a XML (preserving only their changes)
4. User stops posing
5. User starts posing again, with same pose from step 1
6. User can load their 'deltas' from step 3: restoring 'base plus delta' state, netting the same state they saved in step 3 

If we modify base while doing step '2', I'm not sure we'll get the same state back in step '6', as the expected base rotations will be different.

I have wrestled with naming of these components several times. Public/Private, Base/Delta, Diff/Pose... behind the names is the story above, in addition to the requirement to prevent plagiarism or duplication.

I've updated this case with several other notes, which I think https://jira.firestormviewer.org/browse/FIRE-35321 also wants addressed.

Hecklezz: I can test what you said the next time I'm on OS, but what I'm doing is actually restoring the base because it is actually being changed by the visual manipulator and is no longer what it was when you opened the poser.

So what's happening is roughly this:
1. Base scale/position/rotation just say are all 1.f and public scale/position/rotation are all 0.
2. You change the position and scale slider and change the rotation so now base is still all 1.f and say public are all 0.5f.
3. You use the visual manipulator which changes rotation, which now overrides base with your original 0.5f changes and whatever change you made on the visual manipulator, and sets public all back to 0.f
4. You now have changes done in the poser, but all public values are 0.f and all base values are no longer 1.f

Now of course I need to test the steps you said, since what I'm doing could very well still have adverse effects but what happens is this-
1. Base scale/position/rotation just say are all 1.f and public scale/position/rotation are all 0.
2. You change the position and scale slider and change the rotation so now base is still all 1.f and say public are all 0.5f.
3. You use the visual manipulator which changes rotation, which now overrides base with your original 0.5f changes and whatever change you made on the visual manipulator, and sets public all back to 0.f
4. restorePublicState is called which restores the 0.5f public changes and sets base back to what it was (however I now realise a bug which I'll mention below).
5. You now have changes done in the poser and public values are populated and base values are as they were when you started posing.

The bug I just realised with my changes, is the rotational changes made by the visual manipulator are still stored in the baseRotation and not added to the public values.

Hecklezz: If I am understanding the poser, how it functions, and how we want the base and public values to be during the process of using the poser. The "true" fix would be making sure the visual manipulator is writing its changes to the public value and not overriding the joint by adding everything to the base including its own changes and zeroing out public.

AngeldarkRaymaker: Yeah, correct: the base was being changed and the delta went to zero. This introduced a problem with what we would save out (edits from Manipulator would result in no 'delta', so if saved, there was nothing to save for that joint).

There were several issues we had after merging the Poser and Visual Manipulator together; the ones I know of from the poser-side I think I've addressed (though it's not in the public branches yet).

Hecklezz:
This can be closed since the rework appears to have properly fixed this issue.

PR #102, opened 2025-04-03
[FIRE-35317] Restore AudioLevelWind as a debug setting

Hecklezz:
[FIRE-35317](https://jira.firestormviewer.org/browse/FIRE-35317)

LindenLabs a while ago removed the AudioLevelWind debug setting since they considered it "unused". 
This PR simply reverts this change and brings it back, since there are some users who wish to disable to wind sound when hovering still.


beqjanus:
Awesome, thank you

PR #103, opened 2025-04-09
FIRE-34775: For OpenSim switch back to PST/PDT

Tampa:
Based on FIRE-34775, switching back to PST/PDT for the OpenSim version, retaining SLT at SL. I don't have a test environment, but this probably works.

### DIFF
--- a/indra/llcommon/llstring.cpp
+++ b/indra/llcommon/llstring.cpp
@@ -31,6 +31,7 @@
 #include "llfasttimer.h"
 #include "llsd.h"
 #include <vector>
+#include "llviewernetwork.h" // <FS/> Access to GridManager
 
 #if LL_WINDOWS
 #include "llwin32headers.h"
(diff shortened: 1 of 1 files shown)


aiaustin:
I think this is a very sensible change back.  the time zones really are confusing when in OpenSim and given very different daylight savings change over days in different countries.  Using "SLT" in OpenSim simply causes confusion.

beqjanus:
Agree that this makes no sense for OpenSim. That said, I don't think PDT/PST is actually any better, and a longer-term solution should be to provide a reference timezone, perhaps as part of the grid info. 

Checked with LL as to whether the 2009 comment re: deprecation of SLT still stands and they have confirmed that SLT is staying for now (as it gives reliable base for inworld events)

beqjanus:
I'm reverting this is it will not build because it crosslinks headers from other modules (llstring cannot reference llviewernetwork) I'll figure out how to do this properly and take it from there.

Tampa:
The justification from LL for time basis of events is a funny one given the event tab in the legacy search defaults to local time on the system the viewer runs on. With international differences in date change for daylight savings, a time zone that doesn't identify what it currently is set to is questionable, but the worst part is that SLT itself might refer to a completely different time zone. Anyone that would want to know the offset to their local time might get the result for Sri Lanka Time rather than PST equivalent.

Sending a timezone as part of grid info could work. Grid info is static though, so if changes are made to the time zone later on they would not reflect in viewers that already added that grid to their list.

Currently a daylight savings indicator is sent via the login response. That could be extended to include the full time zone, but it would mean breaking login for every viewer until the new response is handled properly.

Ultimately a return to PST/PDT should be enough for now. Thanks :)

PR #104, opened 2025-04-10
Fix build logging when the build path contains spaces

NaomiVexed:
This is just a small change to the configuration script. Previously, tee would interpret build paths containing spaces as two separate arguments. Wrapping the build path in quotes prevents this.


beqjanus:
Thanks for this

PR #105, opened 2025-04-10
Hop protocol fixes and negative Z coordinate OpenSim support

Hecklezz:
Fixes all problems in the two following Jira issues-
https://jira.firestormviewer.org/browse/FIRE-35268
https://jira.firestormviewer.org/browse/FIRE-35333

1. Copying the SLurl on the world map in OpenSim, now correctly creates the SLurl of the grid you are on (you will need to wait a few seconds on a new grid, since the viewer doesn't receive the latest grids simulation features and grid info until a few seconds after arriving. Attempting to immediately copy the SLurl on a new grid will still have the old grid url. There is no  easy way around this without fixing the overarching race condition issue on grid swapping)
2. Hop protocol urls in nearby chat and IM's now get correctly hyperlinked if the coordinates are missing, or if the Z value in the coordinate is negative (supported by OpenSim)
3. The world map now allows you to enter a Z value less than 0 depending on the regions MinSimHeight (This requires an OpenSim server update to fully support, since currently even though building and being at negative Z values are supported, attempting to teleport to one sends you to 128,128,128 until it gets fixed in the near future).

Side fixes related to above changes:
1. Fixed the tracker beacon when it was tracking a coordinate with a negative Z value to no longer underflow due to using U32 and causing a viewer freeze.

### DIFF
--- a/indra/llui/llurlentry.cpp
+++ b/indra/llui/llurlentry.cpp
@@ -1249,7 +1249,7 @@ void LLUrlEntryParcel::processParcelInfo(const LLParcelData& parcel_data)
 //
 LLUrlEntryPlace::LLUrlEntryPlace()
 {
-    mPattern = boost::regex("((hop://[-\\w\\.\\:\\@]+/)|((x-grid-location-info://[-\\w\\.]+/region/)|(secondlife://)))\\S+/?(\\d+/\\d+/\\d+|\\d+/\\d+)/?", // <AW: hop:// protocol>
+    mPattern = boost::regex("((hop://[-\\w\\.\\:\\@]+/)|((x-grid-location-info://[-\\w\\.]+/region/)|(secondlife://)))\\S+(?:/?(-?\\d+/-?\\d+/-?\\d+|-?\\d+/-?\\d+)/?)?", // <AW: hop:// protocol>
                             boost::regex::perl|boost::regex::icase);
     mMenuName = "menu_url_slurl.xml";
     mTooltip = LLTrans::getString("TooltipSLURL");
--- a/indra/newview/llfloaterworldmap.cpp
+++ b/indra/newview/llfloaterworldmap.cpp
@@ -57,6 +57,8 @@
 #include "llnotificationsutil.h"
 #include "llregionhandle.h"
 #include "llscrolllistctrl.h"
+#include "llviewernetwork.h" // <FS/> Access to GridManager
+#include "lfsimfeaturehandler.h" // <FS/> hyperGridURL()
 #include "llslurl.h"
 #include "lltextbox.h"
 #include "lltoolbarview.h"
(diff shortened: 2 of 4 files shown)


beqjanus:
I'm not very happy about -ve Z in general. I swear we've had this argument before. -ve Z is not a supported domain and I have no expectation that this will reliably work in shaders and other aspects of the viewer that treat the coordinate space as +ve only. If this works it works purely by luck and I have no confidence in encouraging the use of this coordinate space when future updates to thte viewer could render it non-functional.

aiaustin:
I understand Beq, and you are probably referring to the discussions on this most recently on the NGC variant Discord channel where they decided to set MinSimHeight at  Z=0m to stay in line with Second Life and potentially avoid any potential issues.  Z = 0m is also the lower limit by default if a MinSimHeight is not provided by the server to the viewer.

But the discussion also took place a while ago for the main dev master branch does allow MinSimHeight to be set and its usually -100m to allow deep water builds while keeping the water level at the default +20m. It actually all works well and would be improved if this PR is incorporated.

The small changes here do make quite a few things "just work" an mostly relate to handling the SLurl hyper links properly i n the few places that are still wonky.  Most people will never notice this and it will just appear to do what they expect, which is not the case at the moment.  Its not really a case of encouraging more usage, just making things that don't work at the moment work as they might be expected to.  

It would be great if you could see your way to incorporate this PR and we take it from there is any issues do arise in future.

beqjanus:
As previously mentioned and for absolute clarity:

TL;DR – I am happy to accept this Pull Request, but doing so should not be interpreted as Firestorm endorsing or committing to support negative Z-coordinate spaces going forward.

Firestorm does not consider negative coordinates (Z or otherwise) as valid. We explicitly do not support negative Z coordinates, and grid owners using negative Z coordinates in OpenSim must be fully aware that this remains unsupported. If this feature breaks in the future, neither I nor the Firestorm team will take responsibility for diagnosing or fixing such issues. Moreover, we will not proactively validate future viewer updates against negative Z coordinate usage.

Historically, requests for negative Z support (e.g., negative Z camera positions) have been explicitly declined due to the unnecessary complexity and maintenance burden involved. Accepting this Pull Request does not alter our position. It simply means we're integrating contributed work beneficial to OpenSim users, like the valuable contributions from Hecklezz. However, if future viewer changes lead to issues—such as rendering glitches, calculation errors, shader problems, or crashes—directly attributable to negative Z coordinate usage, the Firestorm team will not prioritize fixing them.

While we understand the conceptual convenience of defining sea-level at 0, adopting negative Z coordinates is an isolated decision within OpenSim. It inherently introduces potential incompatibilities with future viewer updates. Acceptance of community fixes or contributions regarding negative Z does not imply ongoing support or endorsement by Firestorm.

In summary, negative Z coordinate usage is considered fragile, unsupported, and solely at the risk of the adopting OpenSim grid owners.

PR #106, opened 2025-04-13
Fix GLTF texture panel bugs and inconsistencies compared to SLViewer

Hecklezz:
[FIRE-32413](https://jira.firestormviewer.org/browse/FIRE-32413)
[FIRE-34261](https://jira.firestormviewer.org/browse/FIRE-34261)
[FIRE-34350](https://jira.firestormviewer.org/browse/FIRE-34350)
[FIRE-35138](https://jira.firestormviewer.org/browse/FIRE-35138)
[FIRE-35173](https://jira.firestormviewer.org/browse/FIRE-35173)

While using the legacy LL texture panel instead of the FS alternative one, attempting to edit GLTF settings will cause textures on the material to be reverted/removed.
This was caused by the LLRenderMaterialOverrideFunctor intended for the FS texture panel to be used for the regular one, which resulted in it not applying settings properly. This PR simply renames the functor so it is only used for the FS texture panel and not the LL texture panel.
Additionally, when selecting a texture and clicking the cancel button using the FS texture panel, the chosen texture does not get reverted. This was caused by the FS texture panels LLRenderMaterialOverrideFunctor not implementing reverting changes if a texture selection had been cancelled. This PR reimplements reverting the texture if the selection has been cancelled.
Lastly, when both a PBR material and BP is applied at the same time, viewing the model with the BP tab selected caused the object to display completely incorrectly and parts of the PBR material were applying ontop of BP.
This was caused by the GLTF overrides not being correctly removed when on the BP tab. This PR correctly hides and restores the GLTF material on selected objects when changing between PBR and BP.

### DIFF
--- a/indra/newview/fspanelface.cpp
+++ b/indra/newview/fspanelface.cpp
@@ -243,9 +243,9 @@ private:
 };
 
 //
-// keep LLRenderMaterialOverrideFunctor in sync with llmaterialeditor.cpp just take
-// out the reverting functionality as it makes no real sense with all the texture
-// controls visible for the material at all times.
+// keep LLRenderMaterialOverrideFunctor in sync with llmaterialeditor.cpp just take out the
+// reverting functionality of non texture and color related GLTF settings as it makes no
+// real sense with all the texture controls visible for the material at all times.
 // TODO: look at how to handle local textures, especially when saving materials
 // - Would be nice if we had this in its own file so we could include it from both sides ... -Zi
 //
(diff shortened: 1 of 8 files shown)


beqjanus:
Not yet ready to merge. Tested this evening and not behaving as expected. Sent a video clip to @Hecklezz

beqjanus:
re-tested and now with the extra commit, things work far better. 
Note:
This is not the end of the journey, at least one remaining issue is that the BP show state will persist for the editor after they close the build tools. Things automatically reset once the object has been purged by a restart or enough TPs to flush in-mem caches. 
However this is a million times better than what we have now.

PR #107, opened 2025-04-16
Fix object and HUD clicking in mouselook mode

Sekkmer:
### Description
In the current implementation, clicking objects in mouselook (first-person) mode does not function correctly because the viewer attempts to pick using the hidden mouse cursor location instead of the center of the screen where the crosshair is located. This leads to objects being unclickable, significantly impairing the usability of mouselook mode.

This pull request fixes the issue by adjusting the picking logic within `LLViewerWindow::pickImmediate()` to always use the center coordinates of the viewport when in mouselook mode.

### Implementation Details
Modified `indra/newview/llviewerwindow.cpp`:

```cpp
// <FS:Sek> Pick from center of screen in mouselook
if (gAgentCamera.getCameraMode() == CAMERA_MODE_MOUSELOOK)
{
    x = gViewerWindow->getWorldViewRectScaled().getWidth() / 2;
    y_from_bot = gViewerWindow->getWorldViewRectScaled().getHeight() / 2;
}
// </FS:Sek>
```

### Testing
1. Enter mouselook mode.
2. Attempt to click interactive objects (doors, buttons, scripted objects).
3. Verify clicks now accurately register at the screen's center (crosshair), and object interactions function as expected.
4. Also verify that HUDs can be clicked correctly in mouselook mode (hold Alt).

### Documentation
No additional documentation changes are necessary, as this restores intended functionality.

### Related Issues
This fix addresses a known upstream issue: [[Second Life Viewer Issue #1519](https://github.com/secondlife/viewer/issues/1519)]

### Additional Notes
This fix was tested on Linux. I'm submitting this fix directly to Firestorm because building and testing Firestorm on Linux is significantly easier than the official LL viewer.

### DIFF
--- a/indra/newview/llviewerwindow.cpp
+++ b/indra/newview/llviewerwindow.cpp
@@ -5426,6 +5426,14 @@ LLPickInfo LLViewerWindow::pickImmediate(S32 x, S32 y_from_bot, bool pick_transp
         pick_transparent = true;
     }
 
+    // <FS:Sek> Pick from center of screen in mouselook
+    if (gAgentCamera.getCameraMode() == CAMERA_MODE_MOUSELOOK)
+    {
+        x = gViewerWindow->getWorldViewRectScaled().getWidth() / 2;
+        y_from_bot = gViewerWindow->getWorldViewRectScaled().getHeight() / 2;
+    }
+    // </FS:Sek>
+
     // shortcut queueing in mPicks and just update mLastPick in place
     MASK key_mask = gKeyboard->currentMask(true);
     mLastPick = LLPickInfo(LLCoordGL(x, y_from_bot), key_mask, pick_transparent, pick_rigged, pick_particle, pick_reflection_probe, true, false, NULL);


beqjanus:
Thanks @Sekkmer 
As far as I was aware this issue had been resolved last year (see https://jira.firestormviewer.org/browse/FIRE-34734)
Is this issue a variant of the earlier issue that remained unfixed?

Sekkmer:
> Thanks @Sekkmer As far as I was aware this issue had been resolved last year (see https://jira.firestormviewer.org/browse/FIRE-34734) Is this issue a variant of the earlier issue that remained unfixed?

For me clicking on stuff, like objects, usually fails in FS, while Alt HUD click just makes my cursor jump to the center of the screen, neither of them works reliable for me

I did test it with this change and all issues seem to be resolved by it, then I removed the fix recompiled and they appeared again, and several of my friends had similar issues regarding mouse look where clicking on things doesn't work as expected

beqjanus:
I wonder if this is a Linux only bug as I cannot repro on windows, unless I misunderstand the bug of course. Here's a gyazo of me in mouselook. I can click on lamps the mouseevent is always happening in the right place. I can alt-mouse to drag the UI components and interact with HUDs

https://gyazo.com/9f453d68fb8f5ab374cb11a6e3fcba30

I'll try on Linux in the morning to see if I have a different experience there. 

ooh one thing... please check that you don't have mousewarp disabled. 
https://gyazo.com/3ebf84eb12fdefb6ecdf42a7a3fb3fde

ItsRiumi:
I am on the latest nightly on windows and clicking in mouselook works with some objects, and others for some reason don't.
@beqjanus As far as HUD clicking and dragging, you're probably thinking of the firestorm UI. That works.
But as far as for HUD attachments, the behavior @Sekkmer is describing is what I also get. So no HUD clicking, currently.

beqjanus:
@ItsRiumi linux or windows? and do you have the mousewarp setting disabled?

ItsRiumi:
> @ItsRiumi linux or windows? and do you have the mousewarp setting disabled?

Windows, and yes, I have disabled on false.

beqjanus:
Thanks for this. I get concerned where input and control is involved as it is so easy to break some other input mechanism but as I can't find anyone who is an active mouselook user to test this before I import it, I'm going to import for the next beta build and see where we go from there.

Sekkmer:
I did try to test more, and the bug is seem to be a bit random even on linux, standard beta Firestorm, I had it working and not working and I could not tell any difference since it was basically the same object just some restarts, FS & PC,  in between, however I have not observe any issues why I had the fix, the bug never once appeared on my side and there were no negative regressions either during my tests

PR #110, opened 2025-04-21
[FIRE-35405] Fixes preferences search highlight color by having a color picker for both background and font colors

Hecklezz:
When Firestorm merged in LL's recent changes, it didn't properly account for the preferences search highlight colour being split into background and font colours. This resulted in the colour that you set in preferences to not actually correlate with what the viewer reads for the colour when it highlights searched things, so it defaults to whatever SearchableControlHighlightBgColor and SearchableControlHighlightFontColor are set to in the colors xml.

This can be fix in 2 ways.
1. Revert back to just relying on the 1 colour, this is the simplest and also won't require translation changes.
2. Implement into preferences two colour pickers, one for Background and one for Font. This is less simple, since every language will now need new translations to account for these being split up.

<s>This PR provides a solution using the 1st method since I don't have translations readily available.
If solution 2. is the preferred way to go, I can make the necessary changes to work that way, and then translations will just need to be added.</s>

This PR now provides a solution using the 2nd method. The translations may not be perfect however, so any suggestions you have for any changes, feel free to mention them here for me to change, or push your own changes :)

### DIFF
--- a/indra/llui/llbutton.cpp
+++ b/indra/llui/llbutton.cpp
@@ -901,7 +901,10 @@ void LLButton::draw()
 
     // Highlight if needed
     if( ll::ui::SearchableControl::getHighlighted() )
-        label_color = ll::ui::SearchableControl::getHighlightFontColor();
+        // <FS> [FIRE-35405] Fix for search highlight color until font color picker is added
+        //label_color = ll::ui::SearchableControl::getHighlightFontColor();
+        label_color = ll::ui::SearchableControl::getHighlightBgColor();
+        // </FS>
 
     // overlay with keyboard focus border
     if (hasFocus())
--- a/indra/llui/llbutton.cpp
+++ b/indra/llui/llbutton.cpp
@@ -901,10 +901,7 @@ void LLButton::draw()
 
     // Highlight if needed
     if( ll::ui::SearchableControl::getHighlighted() )
-        // <FS> [FIRE-35405] Fix for search highlight color until font color picker is added
-        //label_color = ll::ui::SearchableControl::getHighlightFontColor();
-        label_color = ll::ui::SearchableControl::getHighlightBgColor();
-        // </FS>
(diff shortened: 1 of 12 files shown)


Ansariel:
#2

Hecklezz:
> #2

Was this your way of saying, Firestorm wants to go the route of the 2nd solution?
Since I assume you didn't mean to link to PR 2.

Hecklezz:
Alright, I have made the necessary changes to allow changing both the background and font colour of the preferences search highlighting.
The translations were done to the best of my ability with translation tools that are available, since I am an English only speaker.

I also did my best to try and fit the changes with all available languages. Some fit better than others, but everything should at least display properly and not be cut-off.

Any suggestions you have for any changes, feel free to mention them here and push your own changes :)

beqjanus:
Thanks! 
I'm going to tag @PanteraPolnocy as she looks after translations. Generally speaking, we leave translations to the translation team as tools don't cut it. I'll let Pantera make a call on that though. 
For future changes, don't worry about the translations; they will get picked up by others, as you did in your first question, it is worth considering the least work option, but if the "right" choice means creating new strings then so be it. The main thing to keep in mind is that we must have unique names to be able to do the translations. Thinking about spacing and being aware that other languages might need more space is always appreciated.

Hecklezz:
I did my best to make absolutely minimal changes to the translations, just adding the word "background" or "font" where the tools deemed appropriate, the rest of their sentence structure and words _should_ be unchanged.
A few of them were left completely unchanged as well since they never even mentioned color at all and I didn't want to have the sentence change too much by adding numerous words.

So I hope they are mostly all good, just the ones that never mentioned colour in the first place are probably the ones that need adjusting if anything.

PanteraPolnocy (approved):
Re: translations. I'd say go ahead this time, but in the future please leave these to the people who maintain their respective translations. Sometimes it's better to leave the string as English than to introduce something completely invalid by accident.

PR #111, opened 2025-04-24
[FIRE-35294] Make minerjr's chat console hiding work with toast chat too

DarlCat:
The feature that minerjr added with FIRE-35039 covered the vintage style console chat, and added the checkbox corner on the chat toolbar icon. It bothered me that this feature didn't work for toast chat since that's my preferred method, so I am providing my implementation that covers this case.


DarlCat:
Pushed a correction that avoids an unnecessary step. After looking at this again a few hours later it occurred to me that calling `LLFloaterIMNearbyChatScreenChannel::removeToastsFromChannel` directly each time the setting is updated shouldn't cause any harm because no toasts would exist to be lost when toggling the visibility of chat console/toasts on from the off state.

PR #113, opened 2025-04-26
Fix mouselook leftclick disabled by passon from `llTakeControl`

Sekkmer:
### Description
This PR reverts #107 and removes `pass_on` checking from `LLAgent::leftButtonGrabbed()` 

The old PR did not address the actual issue, and the `if` condition was always false, so it did nothing and should be removed.

The actual issue is caused by `LLAgent::leftButtonGrabbed()` checks the `TakenPassedOnCount` on top of the `TakenCount` unlike every other `...Grabbed()` function below it

This causes `llTakeControls(CONTROL_ML_LBUTTON, TRUE, TRUE)` to work incorrectly and prevent mouse look click even though it should not do that

### Testing

Here is a minimal LSL script to reproduce and test the issue
Simply placing this into an object and accepting the permission will prevent clicking in mouse look if this fix is not present
(changing the script and removing the `llTakeControl` or adjusting its values will re-enable clicking)

```lsl
default {
    state_entry() {
        llRequestPermissions(llGetOwner(), PERMISSION_TAKE_CONTROLS);
    }
    run_time_permissions(integer perm) {
        if(PERMISSION_TAKE_CONTROLS & perm) {
            llTakeControls(CONTROL_ML_LBUTTON, TRUE, TRUE);              
        }
    }
    control(key id, integer level, integer edge)
    {}
}
```

### Documentation

ref: [Wiki: llTakeControls](https://wiki.secondlife.com/wiki/LlTakeControls)
According to the wiki this restores an intended behavior

### DIFF
--- a/indra/newview/llagent.cpp
+++ b/indra/newview/llagent.cpp
@@ -4140,15 +4140,14 @@ void LLAgent::initOriginGlobal(const LLVector3d &origin_global)
 
 bool LLAgent::leftButtonGrabbed() const
 {
-    const bool camera_mouse_look = gAgentCamera.cameraMouselook();
-    // <FS:Sek> Fix mouse look click when only passed on controls are used
-    return (!camera_mouse_look && mControlsTakenCount[CONTROL_LBUTTON_DOWN_INDEX] > 0)
-        || (camera_mouse_look && mControlsTakenCount[CONTROL_ML_LBUTTON_DOWN_INDEX] > 0);
-    //return (!camera_mouse_look && mControlsTakenCount[CONTROL_LBUTTON_DOWN_INDEX] > 0)
-    //    || (camera_mouse_look && mControlsTakenCount[CONTROL_ML_LBUTTON_DOWN_INDEX] > 0)
-    //    || (!camera_mouse_look && mControlsTakenPassedOnCount[CONTROL_LBUTTON_DOWN_INDEX] > 0)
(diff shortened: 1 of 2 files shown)


Ansariel:
Is this an issue that solely exists on Firestorm or also on the LL viewer?

ItsRiumi:
> Is this an issue that solely exists on Firestorm or also on the LL viewer?

I just tested it on LL viewer and it is present in there as well.

Sekkmer:
> Is this an issue that solely exists on Firestorm or also on the LL viewer?

The code is same in LL viewer [llagent.cpp#L3473](https://github.com/secondlife/viewer/blob/230244b7ea4843dbcefa3bc30862b5b02e178ee7/indra/newview/llagent.cpp#L3473)

But since compiling LL viewer on linux is not ideal, I have no way to test the change on my side, that is why I made the PR for FS since that I can compile without issues on linux

Ansariel:
> > Is this an issue that solely exists on Firestorm or also on the LL viewer?
> 
> I just tested it on LL viewer and it is present in there as well.

Then the PR should be targetted against the LL viewer to maintain a consistent behavior.

Sekkmer:
> > > Is this an issue that solely exists on Firestorm or also on the LL viewer?
> > 
> > 
> > I just tested it on LL viewer and it is present in there as well.
> 
> Then the PR should be targetted against the LL viewer to maintain a consistent behavior.

Sure, I made the PR for LL viewer: [LL #3990](https://github.com/secondlife/viewer/pull/3990)

beqjanus:
As this has been PR'd upstream, I'm going to close this version of the PR to avoid future clashes.

PR #116, opened 2025-05-03
Add support to area search for filtering reflection probes

DarlCat:
This PR adds support to area search for filtering reflection probe objects. The most reliable way I was able to get objects to show be detected was checking for a object coupled to the object, but am open to suggestions for a method that may work better across all circumstances.

Screenshot From 2025-05-02 19-01-25
Screenshot From 2025-05-02 19-01-52

### DIFF
--- a/indra/newview/fsareasearch.cpp
+++ b/indra/newview/fsareasearch.cpp
@@ -156,6 +156,7 @@ FSAreaSearch::FSAreaSearch(const LLSD& key) :
     mFilterPhantom(false),
     mFilterAttachment(false),
     mFilterMoaP(false),
+    mFilterReflectionProbe(false),
     mFilterDistance(false),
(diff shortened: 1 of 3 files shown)


beqjanus:
interesting, I'll have a look at this when I get a moment. Thanks @DarlCat

beqjanus:
@DarlCat what do you think about adding reflection probes to the exclusion list as well. This would be consistent many of the other entries and I can see that this might be useful as probes become more common.

DarlCat:
That's a great idea, I'll see if I can add that to this PR this evening.

DarlCat:
I've added the requested exclusion support. The checkboxes are mutually exclusive and behave as the other exclusion options with the exception that it is not excluded by default.

image

PR #117, opened 2025-05-03
[FIRE-34791] Fix FullScreen failing on Windows 11 24H2

Hecklezz:
[FIRE-34791](https://jira.firestormviewer.org/browse/FIRE-34791)

Possible fix for FullScreen not working on specific Windows setups based on Rilene Starlights information-
https://jira.firestormviewer.org/browse/FIRE-34791?focusedId=246285&#comment-246285
As well as additional information I discovered here-
https://jira.firestormviewer.org/browse/FIRE-34791?focusedId=246321&#comment-246321

This might not fix the issue as a whole, but it should fix the specific issue relating to what Rilene described which I do believe to be the same issue based on information I found.

Duplicate of my changes that got merged into SLV-
https://github.com/secondlife/viewer/pull/4023

### DIFF
--- a/indra/newview/llappviewerwin32.cpp
+++ b/indra/newview/llappviewerwin32.cpp
@@ -1034,6 +1034,29 @@ bool LLAppViewerWin32::reportCrashToBugsplat(void* pExcepInfo)
     return false;
 }
 
+bool LLAppViewerWin32::initWindow()
+{
+    // This is a workaround/hotfix for a change in Windows 11 24H2 (and possibly later)
+    // Where the window width and height need to correctly reflect an available FullScreen size
+    if (gSavedSettings.getBOOL("FullScreen"))
+    {
+        DEVMODE dev_mode;
+        ::ZeroMemory(&dev_mode, sizeof(DEVMODE));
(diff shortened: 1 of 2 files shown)


Ansariel:
This looks more like it needs to go into LLWindowWin32

Ansariel:
That still is Windows API specific code that does not belong in LLAppViewer.

Hecklezz:
Which is why I ifdef'd it... I don't get it, you would want all these other classes to contain an incorrect width and height size of the application window all for the sake of this already ifdef'd code being moved to LLWindowWin32? That seems rediculous to me and sounds like that could introduce more bugs than fix.

Hecklezz:
Unless you mistyped when you mentioned LLWindowWin32 and actually meant LLAppViewerWin32, because yes, having an override and doing this in LLAppViewerWin32 would work and then keep it inside Windows specific code.

Hecklezz:
I have updated the PR and commits that contain just the changes done and merged upstream with SLV.

Ansariel:
Thanks!

PR #118, opened 2025-05-07
[FIRE-35299] Fix world map Find button potentially removing results undesirably

Hecklezz:
[FIRE-35299](https://jira.firestormviewer.org/browse/FIRE-35299)

Fixes the above issue with the same changes recently merged into SLViewer, except also fixed for Starlight and StarlightCUI-
https://github.com/secondlife/viewer/pull/4038


Ansariel:
I'm going to close this PR because it does not fix the actual issue (see my explanation here: https://github.com/secondlife/viewer/pull/4040)

I pushed the correct fix in the meantime: 91abe3643db83f80f3c0386303c344902a8a8169

PR #121, opened 2025-05-12
Remove outdated kfmclient use

SagaVortex:
kfmclient will open all links in konquerer as the browser, ignoring any choices made in KDE settings' Default Application. This causes URL opening to not follow the expected behavior of respecting user set default application.

Searching for kfmclient leads me to believe this is no longer a used tool with KDE: https://techbase.kde.org/Development/Tools/Using_kfmclient If I understand this correctly, the replacement had been to use `kioclient exec` instead of `kfmclient openURL`. Trying this change the script works as intended in testing and opens the correct default browser atleast for me. But I believe that methods like xdg-open will cover more cases and is the more modern approach. So I would propose to skip the special check for KDE entirely.

### DIFF
--- a/indra/newview/linux_tools/launch_url.sh
+++ b/indra/newview/linux_tools/launch_url.sh
@@ -51,12 +51,6 @@ if [ ! -z "$GNOME_DESKTOP_SESSION_ID" ] && which gnome-open >/dev/null; then
     exit
 fi
 
-# Launch the default KDE browser.
-if [ ! -z "$KDE_FULL_SESSION" ] && which kfmclient >/dev/null; then
-    kfmclient openURL "$URL" &
-    exit
-fi
-
 # List of browser commands that will be tried in the order listed. x-www-browser
 # will be tried first, which is a debian alternative.
 BROWSER_COMMANDS="      \


beqjanus:
Hiya, thanks for this. 
Do you happen to know how long this has been out of support within KDE? The link you posed appears to have a date around 2019. The only reason I ask is because I don't want to remove something that might still be a viable build option for some people, but if we are going back to 2019 then I think we're safe to move on.

SagaVortex:
I honestly can't say for sure. I have had no working external browser with firestorm for 2 years and reported the issue several times. I believe Firestorm developers are all using Gnome so always reported 'it works on my system' in response. I have switched to Alchemy and that has been my solution.
Out of sheer curiosity I looked into it just a little bit and found out the issue is simple enough for even someone really not familiar with programming to figure out.

So while I have no clear reference to cite to tell you when this script might have worked as intended, I can say with confidence it has not worked in 2 years for KDE systems.

SagaVortex:
https://github.com/KDE/kde-baseapps/blob/master/konqueror/client/kfmclient.cpp

The kfmclient code itself states: 
```
kfmclient exec is deprecated and kept for compatibility with KDE 3. \n"
                  "            # See kioclient exec for more information.\n").toLocal8Bit
```
and googling KDE4 states it was released in 2008.


I am not sure if using `kioclient exec` hardcoded or using the several methods below in the code is the better direction. I am fine with either but it would be nice to get this resolved <3

SagaVortex:
the alternative solution would be to change the KDE entry into:
```
# Launch the default KDE browser.
if [ ! -z "$KDE_FULL_SESSION" ] && which kioclient >/dev/null; then
    kioclient exec "$URL" &
    exit
fi
```

let me know if you prefer me to commit this solution instead

PR #126, opened 2025-05-17
[FIRE-34494] fixes unable to open an IM with someone who started a group chat

chanayane:
This was an annoying bug! It should be fixed with that PR.

**Summary**
This patch resolves an issue where the viewer incorrectly reused an existing session with the wrong dialog type (e.g., treating a group chat as a one-on-one IM). This led to confusing behavior when opening IMs with users who had previously initiated a group conversation.

**Changes**
_llimview.cpp:_

- Added type-checking logic to LLIMMgr::addSession() to ensure sessions are reused only if their mType matches the requested dialog.
- If a session exists but has a mismatched type, a new session is created instead.
- Used computeSessionID() for deterministic session creation where appropriate; generated a random UUID for conference sessions.
- Logged mismatches with LL_WARNS for debug visibility.

_fsfloaterim.cpp:_

- Patched FSFloaterIM::show() to check the underlying session's type before proceeding.
- Ensured the floater is only reused if the session type matches the intended dialog type.

**Impact**
- Prevents UI confusion caused by reusing session floaters with incorrect context.
- Fixes inability to open a new IM with someone who previously initiated a group chat.
- Improves internal consistency of session management logic.

### DIFF
--- a/indra/newview/fsfloaterim.cpp
+++ b/indra/newview/fsfloaterim.cpp
@@ -1253,6 +1253,22 @@ FSFloaterIM* FSFloaterIM::show(const LLUUID& session_id)
     if (!gIMMgr->hasSession(session_id))
         return nullptr;
 
+    // <AS:chanayane> [FIRE-34494] fixes unable to open an IM with someone who started a group chat
+    // Prevent showing non-IM sessions in FSFloaterIM::show()
+    LLIMModel::LLIMSession* session = LLIMModel::getInstance()->findIMSession(session_id);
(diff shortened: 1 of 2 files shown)


chanayane:
Wait, there's an issue. Cannot click anymore on the message icon on top right of screen, for non private IM messages (like group chats). Will post a fix asap.

chanayane:
I fixed it, I will update the PR soon and give my green light

chanayane:
Okay I fixed it! I think we're good to go :)

PR #129, opened 2025-05-19
Improves GLTF hiding while object is displaying Blinn-Phong

Hecklezz:
Follow up to this PR-
https://github.com/FirestormViewer/phoenix-firestorm/pull/106

This PR further improves the functionality of hiding and reverting the GLTF material while in BP mode by ensuring updates from the server (such as changes made by other users) correctly update the locally saved GLTF materials so they can be reverted properly. Also fixes a bug while making texture changes in BP mode was reapplying the GLTF materials since the message from the server didn't match the locally null'd GLTF materials.

This should be much more consistent than it was before and desync less if other people have made changes to the same objects you have been previously editing in BP mode. 
There is still 1 minor issue, where if two people are editing the same object at the same time, one person changing PBR the other BP, it will cause a desync for the person editing BP, and to fix it you can do either: 1. Change to the PBR tab, unselecting and reselecting the object will resync it. Or 2. Relog.

### DIFF
--- a/indra/newview/fspanelface.cpp
+++ b/indra/newview/fspanelface.cpp
@@ -737,7 +737,6 @@ void FSPanelFace::onMatTabChange()
         if(objectp)
         {
             last_mat = curr_mat;
-            gSavedSettings.setBOOL("FSShowSelectedInBlinnPhong", (curr_mat == MATMEDIA_MATERIAL));
             // Iterate through the linkset and mark each object for update
             for (LLObjectSelection::iterator iter = LLSelectMgr::getInstance()->getSelection()->begin();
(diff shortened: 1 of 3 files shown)


Ansariel:
How does this work with the default (= LL) texture tab layout?

Hecklezz:
Just like originally, this only works with the FS panel, the LL panel still doesn't let you switch to BP if a PBR material is applied.

Ansariel:
> Just like originally, this only works with the FS panel, the LL panel still doesn't let you switch to BP if a PBR material is applied.

Does "still" mean there is a change in one of their future release branches that enables it, or isn't there anything done in regard to that at all?

PR #131, opened 2025-05-20
Small Improvement to Performance Auto Tune feature

Hecklezz:
Currently the Performance Auto Tune feature (located under World->Improve Graphics Speed... or Advanced->Performance Tools->Improve Graphics Speed...) only has the options for Avatar only, or Avatar and Scene, there is no Scene only option. In the SLViewer, there is a Scene only option and the LLPerfStats implementation that Firestorm uses for the feature is already setup to work with Scene only.

This PR simply adds the dropdown option for Scene Only to be in-line with SLViewer and also for those who don't want the feature impacting Avatar Complexity. Another small change the PR makes, is increasing the preferred render distance maximum from 256 all the way up to 1024. Since I see no reason it should be limited to 256 if you are around regions that are easy to render and you could easily render over 256m at a target framerate, and then once you go to a more demanding region, it will dynamically drop it back under anyway.

I also changed a few of the auto tune settings that didn't persist to now persist, to save people the trouble of having to change them every single time you use the feature, which some people may want on all the time and kept it on across logging out and back in.

TODO: Translation request to add this additional combo box value (scene_only) for "FSTuningFPSStrategy" in all the floater_fs_performance.xml translations.

### DIFF
--- a/indra/newview/app_settings/settings_firestorm.xml
+++ b/indra/newview/app_settings/settings_firestorm.xml
@@ -230,6 +230,42 @@
     <key>Value</key>
     <integer>0</integer>
   </map>
+  
+  <key>AutoTuneRenderFarClipMin</key>
+  <map>
+    <key>Comment</key>
+    <string>The lowest draw distance that auto tune is allowed to use</string>
+    <key>Persist</key>
+    <integer>1</integer>
+    <key>Type</key>
+    <string>F32</string>
+    <key>Value</key>
+    <real>32.0</real>
+  </map>
+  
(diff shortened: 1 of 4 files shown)


Hecklezz:
Quick question, since I am changing the persistency of the settings in settings.xml, should these changes be reflected in settings_firestorm.xml instead?
If so, let me know and I'll immediately make the change before this gets merged.

beqjanus:
I don't think autotune is currently fit for purpose in any case. This PR is on hold until we can dive deeper in to it. Basically with all the pipeline changes for pbr the stats are no longer aligned and it needs to be verified once again.

PR #133, opened 2025-05-21
primfeed: cannot use string with constexpr; revert to static const

EricaNebula:
In an effort to be less annoying and more helpful, this reverts the use of `constexpr` where it was causing compilation to fail because `std::string` isn't a literal type.

Using `std::string_view` here would work but then that causes LLSD params to complain because it's expecting a `std::string`, so the path forward seems to be going back to `static const`.

See original comments on the commit:

https://github.com/FirestormViewer/phoenix-firestorm/commit/51c461500dabd9014786b695e9f29f7ed96edab7#r157607816

### DIFF
--- a/indra/newview/fsareasearch.cpp
+++ b/indra/newview/fsareasearch.cpp
@@ -156,6 +156,7 @@ FSAreaSearch::FSAreaSearch(const LLSD& key) :
     mFilterPhantom(false),
     mFilterAttachment(false),
     mFilterMoaP(false),
+    mFilterReflectionProbe(false),
     mFilterDistance(false),
     mFilterDistanceMin(0),
     mFilterDistanceMax(999999),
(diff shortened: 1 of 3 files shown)


Hecklezz:
C++20 did technically add improvements to support constexpr operations with std::string, and C++23 has made further enhancements again, but I suppose the compiler you are using for Linux is missing the functionality, since on my Windows machine, it compiles fine with C++20 support. However, given that this is the only instance in the entire repo that this occurs, it should definitely accommodate compilers that don't yet support it.

EricaNebula:
On Linux I am building with gcc-11.5.0 which only partially implements C++20. I don't think they finished full support until gcc-13. (Everything else on my system compiles with gcc-14, this is the only package I need 11 for.)

minerjr:
We are still targeting gcc-11 with official builds. I also did this goof myself by adding newer c++20 feature and had to change it back. See the link for what c++ features are support by gcc versions.

[https://gcc.gnu.org/projects/cxx-status.html](https://gcc.gnu.org/projects/cxx-status.html)

Windows we are using Visual Studio 2022 and it supports up to c++23. Mac is using a version of Clang which is closer to gcc feature set.

Ansariel:
Maybe we should finally move on from old compiler that aren't supported anymore...

PR #134, opened 2025-05-21
[FIRE-35355] Fixes navigation bar back and forward buttons while hypergridding in OpenSim

Hecklezz:
[FIRE-35355](https://jira.firestormviewer.org/browse/FIRE-35355)
Fixes problem 2 listed in the above issue.

Ever since the existence of the navigation bar, while you are hypergridding in OpenSim, if you try and use the back or forward buttons to return to a different grid than the one you are on, it would simply just fail and error saying it couldn't teleport.

This PR makes it so the navigation bars teleport history stores the slurl of the current region whenever you teleport, then when you initiate a teleport with the back or forward button, if it is a different grid it requests the global position from the server using the stored slurl and teleports to the location on the grid that way (since the global position is dependent on what grid you are on).

### DIFF
--- a/indra/newview/llteleporthistory.cpp
+++ b/indra/newview/llteleporthistory.cpp
@@ -44,6 +44,10 @@
 #include "llavatarname.h"
 #include "llavatarnamecache.h"
 
+#include "llviewernetwork.h" // <FS/> Access to GridManager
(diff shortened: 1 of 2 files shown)


MrEnigmaSr (approved):
Looks GREAT to me :B 👍

PR #137, opened 2025-05-23
Fix preference highlighted text font being applied to chat input field

Hecklezz:
Fixes the issue picked up by Beq where it was changing the text colour of the chat windows input field.

Because of the way LL has the control setup, I needed to break some C++ core guidelines and cast away const to make this work 😅


Hecklezz:
Closing because doesn't account for other edge cases such as checkboxes, needs investigating further later.

PR #139, opened 2025-05-30
Texture Refactor: Smatter bias and texture scaling

minerjr:
This PR is designed to address issues around the scaling of textures and VRAM usage due to recent changes over the past while to texture system.

Added individual per texture bias. When applying bias to on screen textures, Importance to screen can shield textures from having bias applied. Scales based on the bias level. Also modified close to camera calculation to start at 32 meters. As the bias increases, the area around the camera that is protected from the bias will decrease down to 16 meters. Any texture close to camera does not have bias applied. 

Improved how textures are calculated in size so they match the actual size on screen an object is. This fixes large objects far away and small objects up close having different quality of textures applied to them.

Added some new controls for textures in the Preferences->Graphics->Hardware Settings Tab.

image

Auto Scale: On-Screen Textures:
On screen LOD textures will  automatically up/down scale to match the physical size of the object and no longer stick at the highest discard level. Can turn off for the previous method.

Auto-Scale: Off-Screen Textures
This allows the user to force off screen textures to have automatically scale down once off screen.

Near Off-screen Textures: Keep Texture Quality
Off-screen textures that are close to the camera will be treated as if they are on screen and close to the camera. This protects them also from bias being applied to them.

Moving Fast/Turning: Lower Texture Quality for Faster Speed 
While moving forward fast, or doing fast turns, all textures will have a reduction in quality to try to make loading them faster. This is the existing behaviour, now exposed to allow the user to turn on/off.

Removed artificial limit on VRAM calculations. VRAM that appeared in the viewer and calculations was actually 1/2 the calculated amount, due to wanting to give more wiggle room . The latest version from LL introduced a new divider to further slash the size of the VRAM, so there is no need to double dip on reducing the amount of VRAM the viewer can use. Just have the 1 new control select the divider for the VRAM...

Added new checks to only downscale a texture after 3 consecutive requests for lower quality. Any check that goes back up in quality will reset the counter.

Added option to skip waiting for fetches to finish before going on the next texture. This is more helpful for fetches that take a long time due to slow internet download speed or issues with slow writes to disk or slow read times. It returns and will get the fetch on the next update cycle. There is a debug flag to turn this off called "FSFastMoveLowText". Didn't expose it for now.

Removed forcing all textures to be updated when ever the Bias first goes over 1.00. Depending on the # of textures loaded would cause large lag spike on the users as they were all updated in a single frame.

Reduced amount of time deleting objects per frame from 10 ms to 1 ms. Takes 10x longer, but smooths out the frames more.

Fixed scale down edge case which would try to downscale textures beyond mMaxDiscardLevel of an texture.

Added updated decode priority value for fetched textures so that the higher the boost level, the higher on the priority fetch queue.

Fixes some issues with incorrect checks being done. As well as refactored to code to remove unneeded changes and revert code back closer to original LL code to make adding changes easier. Will eventually push a PR to LL with changes.

Added additional Importance to Camera and In Frustum to textures allow for debug output in the Developer Menu Dump.

Cleaned up and updated logic for the LLViewerTextureList::updateImageDecodePriority to be up to date with latest LL version with changes reapplied.

Removed changes to llcamera as they were no longer needed.

### DIFF
--- a/indra/newview/llviewertexture.cpp
+++ b/indra/newview/llviewertexture.cpp
@@ -90,6 +90,8 @@ S32 LLViewerTexture::sRawCount = 0;
 S32 LLViewerTexture::sAuxCount = 0;
 LLFrameTimer LLViewerTexture::sEvaluationTimer;
 F32 LLViewerTexture::sDesiredDiscardBias = 0.f;
+F32 LLViewerTexture::sPrevDesiredDiscardBias = 0.f;  // <FS:minerjr> Previous desired discard bias
+constexpr S32 NUM_FRAMES_SCALE_DOWN_TEXTURE = 3; // <FS:minerjr> Store number of frames to scale down texture
 U32 LLViewerTexture::sBiasTexturesUpdated = 0;
 
 S32 LLViewerTexture::sMaxSculptRez = 128; //max sculpt image size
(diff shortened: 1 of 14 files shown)


indra/newview/llviewertexture.cpp
```diff
@@ -550,7 +560,7 @@ void LLViewerTexture::updateClass()
     {
         // slam to 1.5 bias the moment we hit low memory (discards off screen textures immediately)
         sDesiredDiscardBias = llmax(sDesiredDiscardBias, 1.5f);
-
+#if 0 // <FS:minerjr> This would cause large hitch if the bias was turned on, and if the bias would ride back and forth, it could keep adding many updates to the texture list
```
Ansariel, line 563: Don't you rather want a hiccup in fps than a viewer that crashes because it ran out of memory?

indra/newview/llviewertexturelist.cpp
```diff
@@ -1095,45 +1098,37 @@ void LLViewerTextureList::updateImageDecodePriority(LLViewerFetchedTexture* imag
...
+        static LLCachedControl<bool> auto_scale_on_screen_textures(gSavedSettings, "FSAutoOnScrText", true);
+        static LLCachedControl<bool> auto_scale_off_screen_textures(gSavedSettings, "FSAutoOfScrText", true);
+        static LLCachedControl<bool> exclude_off_screen_close_to_camera(gSavedSettings, "FSExcludeOffScreenCtoC", true);        
+        // </FS:minerjr>
 
         F32 max_vsize = 0.f;
         bool on_screen = false;
 
         U32 face_count = 0;
+        U32 max_faces_to_check = 1024;
```
Ansariel, line 1111: Isn't there something missing here? Max faces is always 1024. See https://github.com/secondlife/viewer/commit/89512d44f82dcb9679067bb2303acc40b7b43951 - That would make it the same as in LL's `develop` branch.

PR #144, opened 2025-06-09
Added buttons on sounds explorer window to mute all attached / rezzed sounds by avatar

Gaalrauch:
Added buttons on sounds explorer window to mute all attached / rezzed sounds by avatar

Screen1
Screen2

Just in case: I did not update XML files for all langages, only EN

(sorry for multiple posts, I'm not a git master)

### DIFF
--- a/indra/newview/NACLfloaterexploresounds.cpp
+++ b/indra/newview/NACLfloaterexploresounds.cpp
@@ -75,9 +75,9 @@ bool NACLFloaterExploreSounds::postBuild()
     getChild<LLButton>("stop_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::handleStop, this));
     getChild<LLButton>("bl_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::blacklistSound, this));
     getChild<LLButton>("stop_locally_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::handleStopLocally, this));
-    getChild<LLButton>("block_avatar_attached_sounds_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::blacklistAvatarSoundsByFlagWorn, this));
-    getChild<LLButton>("block_avatar_rezzed_sounds_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::blacklistAvatarSoundsByFlagRezzed, this));
-    getChild<LLButton>("block_avatar_gesture_sounds_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::blacklistAvatarSoundsByFlagGesture, this));
+    getChild<LLButton>("block_avatar_worn_sounds_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::blacklistAvatarSoundsByFlag, this, FSAssetBlacklist::eBlacklistFlag::WORN));
+    getChild<LLButton>("block_avatar_rezzed_sounds_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::blacklistAvatarSoundsByFlag, this, FSAssetBlacklist::eBlacklistFlag::REZZED));
+    getChild<LLButton>("block_avatar_gesture_sounds_btn")->setClickedCallback(boost::bind(&NACLFloaterExploreSounds::blacklistAvatarSoundsByFlag, this, FSAssetBlacklist::eBlacklistFlag::GESTURE));
 
     mHistoryScroller = getChild<LLScrollListCtrl>("sound_list");
     mHistoryScroller->setCommitCallback(boost::bind(&NACLFloaterExploreSounds::handleSelection, this));
@@ -101,7 +101,7 @@ void NACLFloaterExploreSounds::handleSelection()
     childSetEnabled("play_locally_btn", num_selected);
     childSetEnabled("stop_btn", num_selected);
     childSetEnabled("bl_btn", num_selected);
-    childSetEnabled("block_avatar_attached_sounds_btn", num_selected);
+    childSetEnabled("block_avatar_worn_sounds_btn", num_selected);
     childSetEnabled("block_avatar_rezzed_sounds_btn", num_selected);
     childSetEnabled("block_avatar_gesture_sounds_btn", num_selected);
 }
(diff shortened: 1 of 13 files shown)


Gaalrauch:
Hi, thanks. I'm sorry about all those "random" lines and in fact I forget to remove a temp comment I made in french xD. I made a version with 3 types, attached, rezzed, and gesture sounds. You may be interested by this one ?

Ansariel:
I am trying to find the actual usecase for this, because:

- If you find a sound annoying in general, you probably would block it in general - or are they situations in which you like to hear that sound?
- On the other hand, if it is a normal sound but somebody is walking around spamming it, it's most likely a troll/griefer and you would block that person anyway.

This seems to be a very niche usecase (if it exists) and introducing 3 new asset types just for that seems overkill - not to mention the problem that arises if LL adds a new asset type as well, which then results in having to change the asset type for these blacklist asset types, which also means there would be a conversion of the saved blacklisted assets.

Gaalrauch:
For now, I’ve implemented eBlacklistFlag as: 

    enum class eBlacklistFlag
    {
        NONE = 0,
        WORN,
        REZZED,
        GESTURE
    };
 
This allows the same avatar to appear up to three times in the blacklist window, with three different types, like this: 
image

Currently, removing one of these three entries also removes the others.

I haven’t yet implemented combined values (bitflags) as suggested by Ansariel’s original design of eBlacklistFlag, but I might revisit that depending on how useful or necessary it becomes.

indra/newview/NACLfloaterexploresounds.cpp
```diff
@@ -513,3 +519,159 @@ void NACLFloaterExploreSounds::onBlacklistAvatarNameCacheCallback(const LLUUID&
     }
     FSAssetBlacklist::getInstance()->addNewItemToBlacklist(asset_id, av_name.getCompleteName(), region_name, LLAssetType::AT_SOUND);
 }
+
+// add avatar attachments sounds to blacklist
+void NACLFloaterExploreSounds::blacklistAvatarSoundsByFlagWorn()
```
Ansariel, line 524: You only need 2 methods here in total:

- blacklistAvatarSoundsByFlag
- onBlacklistAvatarSoundsByFlagNameCacheCallback

The flag enum value can be passed in/through in the boost::bind calls.

indra/newview/fsassetblacklist.h
```diff
@@ -46,11 +46,20 @@ class FSAssetBlacklist : public LLSingleton<FSAssetBlacklist>
 public:
     void init();
     bool isBlacklisted(const LLUUID& id, LLAssetType::EType type);
-    void addNewItemToBlacklist(const LLUUID& id, const std::string& name, const std::string& region, LLAssetType::EType type, bool permanent = true, bool save = true);
+    void addNewItemToBlacklist(const LLUUID& id, const std::string& name, const std::string& region, LLAssetType::EType type,bool permanent = true, bool save = true);
     void addNewItemToBlacklistData(const LLUUID& id, const LLSD& data, bool save = true);
     void removeItemFromBlacklist(const LLUUID& id);
     void removeItemsFromBlacklist(const uuid_vec_t& ids);
     void saveBlacklist();
+    enum class eBlacklistFlag
```
Ansariel, line 54: As mentioned in my previous comment, I would prefer having the enum setup as bitfield so you can have multiple combinations of flags set/checked at the same time.

Gaalrauch:
Merged the previously separate blacklist handling for WORN, REZZED, and GESTURE into a unified system using flags.
Refactored eBlacklistFlag to behave as a bitfield enum, enabling combinations like WORN | GESTURE.
Updated existing blacklist-related functions (such as isBlacklisted(), addNewItemToBlacklist(), etc.) to accept an eBlacklistFlag parameter.
Implemented flag-aware logic in isBlacklisted() so that checks only return true if the specified flag(s) are present.
Modified addNewItemToBlacklist() to:
    Add a new flag to an existing entry if the UUID is already blacklisted.
    Create a new entry if none exists.
Improved removeFlagFromItem() to remove individual flags while keeping the remaining ones intact, and delete the entry only when no flags remain.
Adjusted the UI to display one list entry per active flag.
Ensured list updates are clean and avoid duplication by:
    Removing existing entries before re-adding updated flag sets.
    Deferring reconstruction when multiple flags are being removed in batch.

Ansariel:
This looks already better. It probably would be better to just add another column which shows the flag value instead of reusing the "type" column (the long texts don't fit in there by default anyway).

My original idea was to have one entry for the asset and then have a column for the flags, showing them all in just a single string. That however would have required a bunch of extra logic to just remove one of the flags, so... Maybe we leave that for now. However, adding a new column which shows the label for each flag would be nice instead of re-using the "type" column. Then you don't need that huge chunk of duplicated code in `FSFloaterAssetBlacklist::addElementToList`

Other than that it looks fine to me.

indra/newview/fsassetblacklist.cpp
```diff
@@ -87,29 +87,72 @@ bool FSAssetBlacklist::isBlacklisted(const LLUUID& id, LLAssetType::EType type)
...
+            data["asset_region"] = region;
+            data["asset_date"] = input_date;
+            data["asset_permanent"] = permanent;
+
+            addNewItemToBlacklistData(id, data, save);
+        }
+    }
+    else
+    {
+        // Nouvelle entr�e
```
Ansariel, line 146: English comment please

Gaalrauch: Oops

Gaalrauch:
I'm wondering whether it might be better to replace the use of AT_SOUND with AT_PERSON for these features. We're not really muting a sound itself, but rather something that belongs to an avatar. Also, should we consider renaming the flags if we’re planning for additional use cases? For example:

enum class eBlacklistFlag
{
    NONE    = 0,
    MUTE_AVATAR_WORN_OBJECTS_SOUNDS    = 1 << 0,
    MUTE_AVATAR_REZZED_OBJECTS_SOUNDS  = 1 << 1,
    MUTE_AVATAR_GESTURES_SOUNDS        = 1 << 2
};

Ansariel:
You are blacklisting a sound - blacklisting a person means not rendering it at all.

Could you please address this:

> It probably would be better to just add another column which shows the flag value instead of reusing the "type" column (the long texts don't fit in there by default anyway). That makes it more general for possible future extensions.

Gaalrauch:
Done, I removed the "huge chunk of duplicated code" in FSFloaterAssetBlacklist::addElementToList too

Ansariel:
I'll merge this - then we'll see how it goes.

chanayane:
Nevermind, a full rebuild fixed it. Sorry.

PR #145, opened 2025-06-16
[FIRE-35656] Add privacy option to block payment notification from mu…

casperwardensl:
This PR adds a privacy option t make payment notifications not show at all if the payment is from muted avatars.

The current implementation removes the "reason", but users can still be harassed by the dialogs - it can be distressing to see a name pop up continuously in certain situations.


beqjanus:
@casperwardensl poke again.

PR #151, opened 2025-07-26
[FIRE-35804] Implemented seperate Conversations opacity options

Hecklezz:
Added a new slider under Preferences->Colors->Miscellaneous to give control over the opacity of specifically the conversations floater.
Also added a checkbox below it that allows the conversation floater to abide by the active opacity when it is focused, so that way it can have its own inactive transparency separate to everything else.

Need to fix before merge:
* ~~The Group and Friend filter editors for some reason don't abide by the opacity. I tried overriding their getCurrentTransparency but it seemingly doesn't get called.~~ Fixed


indra/llui/lluictrl.cpp
```diff
@@ -1086,6 +1086,14 @@ F32 LLUICtrl::getCurrentTransparency()
 
     }
 
+    // <FS:TJ> [FIRE-35804] Allow the IM floater to have separate transparency
+    // This is specifically for making the text editors such as chat_editor always active opacity when the IM floater is focused
+    if (mGetIMOpacityCallback && mTransparencyType != TT_ACTIVE)
+    {
+        return llmin(mGetIMOpacityCallback(), alpha);
```
Ansariel, line 1093: Shouldn't this be llmax instead?

indra/newview/app_settings/settings.xml
```diff
@@ -26707,5 +26707,27 @@ Change of this parameter will affect the layout of buttons in notification toast
...
+    <key>FSImActiveOpacityOverride</key>
+    <map>
+      <key>Comment</key>
+      <string>When enabled, uses the Active Opacity value when IM window is focused</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>Boolean</string>
+      <key>Value</key>
+      <real>0</real>
```
Ansariel, line 26730: `<integer>0</integer>`

indra/newview/fsfloaterimcontainer.cpp
```diff
@@ -373,6 +373,22 @@ FSFloaterIMContainer* FSFloaterIMContainer::getInstance()
...
+    static LLCachedControl<F32> im_opacity(gSavedSettings, "FSIMOpacity", 1.0f);
+    static LLCachedControl<bool> im_active_opacity_override(gSavedSettings, "FSImActiveOpacityOverride", false);
+
+    F32 floater_opacity = LLUICtrl::getCurrentTransparency();
+    if (im_active_opacity_override && getTransparencyType() == TT_ACTIVE)
+    {
+        return floater_opacity;
+    }
+
+    return llmin(im_opacity(), floater_opacity);
```
Ansariel, line 388: Probably needs to be llmax here as well?

indra/llui/lluictrl.cpp
Hecklezz: That way the global inactive value takes priority for those who don't have the conversations slider less than the inactive slider (such as people who have no intent on changing this new slider) 

If the override always applied, then the chat bar will always be the opacity of the conversations slider, even if the window isn't focused.
So say your active slider is 1, inactive slider is 0.5, and conversations is at 0.75. When the conversations isn't focused, it would use the inactive value of 0.5, but the chat bar itself would still be 0.75 (the callback is used on the chat bar because overriding opacity on the I'm container does it to almost everything except the chat bar).

Ansariel: That's not the point I was making: If you have a callback, use the value from the callback, don't do a llmin - it will only override the opacity in one way which is not obvious.

Hecklezz: I can share a with and without llmin screenshot if you like to show why it's necessary?

Hecklezz: Unless you are suggesting adjusting the callback to take in the current transparency state, the alpha, and doing the llmin inside there if needed, a s returning the final value, and returning that?
If so then yes, I understand you now and can do that.

Hecklezz: 1. Alright I understand what you mean now
2. Because LLChatEntry is used in notifications list items, and rlv floaters apparently as well from what I saw, so additional code would need to be added to check is the LLChatEntry in the conversations. Also because I'll need to do it for the Filter bars too which aren't LLChatEntry's, so instead of having overrides in multiple places which also need to check if they are from the conversations floater, just set a callback in places that need it, and have one place (LLUICtrl) call the callback. That was my logic anyway for doing it this way.
3. I simply did this to not have an identical function in both, but you make a good point, I can change that easily.

Hecklezz: I'll resolve this conversation and re-request a review so it isn't so messy and you can let me know if you are more happy with the changed implementation and more generic callback :)
If not, then I'm very happy to continue to receive feedback! ^_^

Ansariel (approved):
Seems good to me.

BTW: Particularly for LLUI, there is LLUICachedControl which gives you access to all settings.

PR #153, opened 2025-08-10
Add support for memory map and 64 bit files

minerjr:
This PR adds support for APR memory map files as well as adds 64 bit version of file access methods to allow for files larger then 2 GB from being used. Memory map files are fixed in size (requires file to be re-opened if size changes). Does not affect existing code as separate named methods are used and original methods are left alone.

Added support for 64 bit files
Added memory map file support
Added method to initialize fixed size memory map file 
Change file size limit from 32bit to 64bit (2 GB to 8EB) 
Updated LL_ARP_X defines to match newer #defines in source APR library as older defines are deprecated and give warnings

### DIFF
--- a/indra/llcommon/llapr.cpp
+++ b/indra/llcommon/llapr.cpp
@@ -317,17 +317,38 @@ private:
 //
 LLAPRFile::LLAPRFile()
     : mFile(NULL),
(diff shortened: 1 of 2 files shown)


minerjr:
Going to submit the LL instead.

PR #160, opened 2025-09-03
fix kakadu version in autobuild.xml

chanayane:
Fixed the Kakadu version in autobuild.xml (8.4.1 -> 8.5.0)

I will try to build with the new 8.6.0 soon.

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -1367,7 +1367,7 @@
         <key>copyright</key>
         <string>Kakadu software</string>
         <key>version</key>
-        <string>8.4.1</string>
+        <string>8.5.0</string>
         <key>name</key>
         <string>kdu</string>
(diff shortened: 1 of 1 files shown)


beqjanus:
I'll get the new 8.6.0 official package done tomorrow.

Ansariel:
This has no effect on functionality at all. The version string isn't even used anywhere in the process.

chanayane:
> This has no effect on functionality at all. The version string isn't even used anywhere in the process.

Then should we remove all the version strings from autobuild.xml? Only keeping what's actually useful. Or is it required for Autobuild to work, maybe?

I wish Autobuild could use YAML or even JSON instead of XML. That format is really too easy to screw up and way too verbose. Same goes for XUI.

PR #197, opened 2026-02-20
FIX - Profile Pick save buttons wrongfully enabled while preview mode is active

trish-sl:
I could not find a JIRA referencing this problem.

## Bug
When using the preview mode within profile picks, the save button is correctly disabled.
However, several actions can cause this to be bypassed. When saving a pick, mPreview mode should be OFF or turned off to prevent the pick from saving while in preview mode. If a pick is saved while in preview, it saves incorrectly (it removes links).

## Reproduction
1) A first case of this bug is when you edit a pick with a link inside, go to preview mode, then close the floater. The pop up that asks you to save or discard the changes will cause the pick to save while in preview mode if you click "Save". Instead, it should turn off preview mode, then save before closing the floater.
2) A second case of this bug is when you edit a pick, go to preview mode, then "Set location". The save buttons are wrongfully re-enabled.

## Fix
For case 1), I've made it so the save & quit will toggle off preview before saving.
For case 2), I gated the function to enable save buttons to cover all cases of it being enabled while preview is on. This is the least invasive change I could find that will remain compatible with future changes from LL

### DIFF
--- a/indra/newview/llpanelprofile.cpp
+++ b/indra/newview/llpanelprofile.cpp
@@ -2301,6 +2301,13 @@ void LLPanelProfileSecondLife::onHideAgeCallback()
 
 void LLPanelProfileSecondLife::onSaveDescriptionChanges()
 {
+    // <FS:Trish> Fix applying changes when exiting profile causing the preview text to apply if preview mode is active
+    if (mPreview)
+    {
+        onCommitMenu(LLSD("preview"));
+    }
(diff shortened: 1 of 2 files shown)


trish-sl:
Turns out, the same bug was happening in the profile description and first life tab, so I incorporated the same fix there. (Turn off preview mode after choosing save & quit)

PR #223, opened 2026-05-03
DRAFT: Add mechanism to de-risk and warn about suspicious links in IMs

emmelinedevere:
# Overview

I think the viewer can do more to reduce the harm of phishing scams that we've seen in group messages lately. This is one approach, matching on some common deceptive patterns and commonly-used free app hosts, and making them less likely to be clicked without hiding them entirely.

# Visual Display of Phishing Links

A message with a suspicious URL in it would pop up to the recipient in red, with `[PHISHING?]` prepended to it:

phish2

And in the IM chat window the link would have the same warning prepended, and not be clickable. (I couldn't figure out how to make it red there.) This means that for the attack to succeed, the victim has to actively copy/paste the URL.

phish3

The behavior can be disabled by un-checking this box:

phish4

# Detecting Phishing Links

The solution contained in this PR uses a scoring system with a few signals to infer phishing-ness.

1. **Is the link trying to impersonate secondlife.com or marketplace.secondlife.com?** (I've also thrown in maps.secondlife.com as that's the other kind of official link that people share a lot.) Here we use a Levenshtein distance comparison of the beginning of the domain name. Emphasis is placed mostly on the beginning, as that's what users are most likely to see before they click. There's also a regex-based piece to this (which is what I started with), but it could arguably be dropped if the less-brittle distance comparison is doing its job. I'm also considering switching to Jaro Winkler distance, as that automatically adds weight to similarity at the beginning of the compared strings. If the scammers want to evade this, they'll have to give up some of the effectiveness of their URLs.
2. **Is the domain on a free PaaS or dirt cheap TLD?** There is a finite number of these in the world. They come and go, but not every day.

# Q&A

## Doesn't this just start us down a cat and mouse game that we can never win?

"Winning" is not a binary. Much of security comes down to reducing the likelihood of a successful attack and raising the costs of the attackers. There are only so many free app hosts out there, and only so many ways to mis-spell "marketplace" and "secondlife" before your phishing URL stops looking believable. We can't catch all scammy links with a hard-coded solution, but we can catch most of them that are impersonating `secondlife.com` and `marketplace.secondlife.com`.

## Shouldn't this be done on the server?

Yeah, it should be. Linden Labs could scan all IMs for URLs and run them against something like the [Google Safe Browsing API](https://developers.google.com/safe-browsing/reference/rpc) before anyone sees the message. My testing has shown they're already doing *something* like that, because some of the old phishing links in my chat logs that I've tried to send to my alt ended up with the message getting dropped before my alt could even see the message. On the other hand, a lot of the links from old logs still work, even ones that Google identifies as unsafe at https://transparencyreport.google.com/safe-browsing/search .

## Shouldn't we put the patterns in a database?

Yeah, probably. Scammers will probably always be faster to switch tactics than Firestorm developers want to push out new releases.

In addition to the domain prefixes that this looks for (secondlife.com, marketplace), there are other pretty-commonly-shared domains in SL that scammers would likely hop to next (gyazo, primfeed, flickr), though they're less juicy targets monetarily. Putting those prefixes in an online database or even a text file on Github would make sense, along with the list of free PaaSes and sketchy TLDs.

### DIFF
--- a/indra/llui/llphishingfilter.h
+++ b/indra/llui/llphishingfilter.h
@@ -1,0 +1,66 @@
+/**
+ * @file llphishingfilter.h
+ * @brief Phishing link detection and mitigation.
+ *
+ * $LicenseInfo:firstyear=2026&license=viewerlgpl$
+ * Second Life Viewer Source Code
+ * Copyright (C) 2026, Phoenix Firestorm Project, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation;
+ * version 2.1 of the License only.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * Phoenix Firestorm Project, Inc., 945 Battery Street, San Francisco, CA  94111  USA
+ * $/LicenseInfo$
+ */
+
+#ifndef LL_PHISHINGFILTER_H
+#define LL_PHISHINGFILTER_H
+
+#include "llsingleton.h"
+#include <string>
+#include <vector>
+
+class LLPhishingFilter : public LLSingleton<LLPhishingFilter>
+{
+	LLSINGLETON(LLPhishingFilter);
+	~LLPhishingFilter();
+
+public:
+	/**
+	 * @brief Evaluates the risk score of a given URL.
+	 * @param url The URL to evaluate.
+	 * @return A score between 0 and 100, where higher is more suspicious.
+	 */
+	S32 evaluateURLRisk(const std::string& url) const;
+
+	/**
+	 * @brief Checks if a URL is considered suspicious based on its risk score.
+	 * @param url The URL to check.
+	 * @return True if the URL is suspicious.
+	 */
+	bool isSuspicious(const std::string& url) const;
+
+	/**
+	 * @brief Processes a message and wraps it in a warning if phishing is detected.
+	 * @param message The original message text.
+	 * @param filtered_message The resulting message (with warning if necessary).
+	 * @return True if phishing was detected and the message was altered.
+	 */
+	bool processMessage(const std::string& message, std::string& filtered_message) const;
+
+private:
+	std::string getBaseDomain(const std::string& url) const;
+};
+
+#endif // LL_PHISHINGFILTER_H
--- a/indra/llui/llphishingfilter.h
+++ b/indra/llui/llphishingfilter.h
@@ -51,14 +51,6 @@ public:
 	 */
 	bool isSuspicious(const std::string& url) const;
 
-	/**
-	 * @brief Processes a message and wraps it in a warning if phishing is detected.
-	 * @param message The original message text.
-	 * @param filtered_message The resulting message (with warning if necessary).
-	 * @return True if phishing was detected and the message was altered.
-	 */
-	bool processMessage(const std::string& message, std::string& filtered_message) const;
-
 private:
 	std::string getBaseDomain(const std::string& url) const;
 };
--- a/indra/llui/llphishingfilter.h
+++ b/indra/llui/llphishingfilter.h
@@ -54,6 +54,7 @@ public:
 private:
 	std::string getBaseDomain(const std::string& url) const;
 	std::string getHostname(const std::string& url) const;
+	size_t calculateLevenshteinDistance(const std::string& s1, const std::string& s2) const;
 };
 
 #endif // LL_PHISHINGFILTER_H
--- a/indra/llui/llphishingfilter.h
+++ b/indra/llui/llphishingfilter.h
@@ -53,6 +53,7 @@ public:
 
 private:
 	std::string getBaseDomain(const std::string& url) const;
+	std::string getHostname(const std::string& url) const;
 };
 
 #endif // LL_PHISHINGFILTER_H
--- a/indra/llui/llphishingfilter.h
+++ b/indra/llui/llphishingfilter.h
@@ -40,7 +40,7 @@ public:
 	/**
 	 * @brief Evaluates the risk score of a given URL.
 	 * @param url The URL to evaluate.
-	 * @return A score between 0 and 100, where higher is more suspicious.
+	 * @return A score between 0 and 100+, where higher is more suspicious.
 	 */
 	S32 evaluateURLRisk(const std::string& url) const;
 
@@ -51,10 +51,15 @@ public:
 	 */
 	bool isSuspicious(const std::string& url) const;
 
+	// Scoring components broken out for individual testing
+	S32 getRegexScore(const std::string& hostname) const;
+	S32 getLevenshteinScore(const std::string& hostname) const;
+	S32 getProviderTLDScore(const std::string& hostname, const std::string& domain) const;
+	size_t calculateLevenshteinDistance(const std::string& s1, const std::string& s2) const;
+
 private:
 	std::string getBaseDomain(const std::string& url) const;
 	std::string getHostname(const std::string& url) const;
-	size_t calculateLevenshteinDistance(const std::string& s1, const std::string& s2) const;
 };
 
 #endif // LL_PHISHINGFILTER_H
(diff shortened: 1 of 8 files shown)


Hecklezz:
One big issue this introduces, is us being blamed for people being scammed due to a mechanism like this not being perfect and always able to detect phishing links, and therefore in peoples minds it is our problem for not saying the link was bad that they clicked on.
Or the inverse can also be an issue, where a link it detects as a possible phishing link when it isn't, and then cause people to be falsely reported and/or banned from places for a false positive and then blame us saying they got banned as our system said they tried to possibly phish people.

It is part of the same reason why the Omnifilter in Firestorm has no built in filters for spam/phishing links which could do quite a lot of what this PR is trying to accomplish, because we do not want any responsibility for being the ones saying what is or isn't unwanted text or links, and having to make sure it is always accurate and up-to-date and false-positive free.

But this is just my opinion and I know at least one other on the Firestorm team agrees, but we will discuss to see if it is something we want in Firestorm or not. Thank you!

emmelinedevere:
Thanks for the comment. Those are tradeoffs worth thinking about (e.g. how much harm is being done from people clicking these links today vs how much harm would be done to Firestorm from people complaining about the flagging not being perfect?)

Now that you mentioned it I remember seeing the word "omnifilter" in Beq's recent blog post but didn't connect it to this (and I still need to look into how it works). Does omnifilter hide the message entirely?

I think flagging the message and making the link unclickable ends up putting less of the responsibility on the Firestorm devs than if messages were hidden entirely. The downside of false positives is decreased (people can still copy/paste). If someone is abuse reported to a sim owner, they can appeal and have someone knowledgeable look into it. And if someone still copy/pastes a phishing URL into their browser... well there's only so much we can do. :)

echoboxfox:
I would love this in any form, links are getting dangerous!

Could do it in a form with 
[x] Allow known safe links (whitelist)
[x] Allow links that are not obviously bad (this patch)
[ ] Allow all links

indra/newview/app_settings/settings_firestorm.xml
```diff
@@ -49,6 +49,18 @@
       <integer>0</integer>
     </map>
 
+    <key>FSPhishingFilterEnabled</key>
```
Ansariel, line 52: This needs to be added to `settings.xml` instead - `settings_firestorm.xml` is only an override for a particular mode.

indra/newview/llimview.cpp
```diff
@@ -65,6 +65,7 @@
 #include "llviewerwindow.h"
 #include "llnotifications.h"
 #include "llnotificationsutil.h"
+#include "llphishingfilter.h"
```
Ansariel, line 68: Unnecessary as nothing is used here directly.

indra/llui/lltextbase.cpp
```diff
@@ -2699,18 +2716,30 @@ void LLTextBase::appendTextImpl(const std::string& new_text, const LLStyle::Para
                 setLastSegmentToolTip(match.getTooltip());
             }
 
+            // <FS:Emme> Phishing filter: skip query part coloring/mangling if it's phishing
+            if (is_phishing)
+            {
+                setLastSegmentToolTip("[WARNING: LIKELY PHISHING ATTEMPT]");
```
Ansariel, line 2722: This needs to be made localizable.

indra/llui/llphishingfilter.h
```diff
@@ -0,0 +1,65 @@
...
+	LLSINGLETON(LLPhishingFilter);
+	~LLPhishingFilter();
+
+public:
+	/**
+	 * @brief Evaluates the risk score of a given URL.
+	 * @param url The URL to evaluate.
+	 * @return A score between 0 and 100+, where higher is more suspicious.
+	 */
+	S32 evaluateURLRisk(const std::string& url) const;
```
Ansariel, line 45: The methods defined here should preferably use `std::string_view` instead of `std::string` const-refs to reduce memory allocations.

indra/llui/llphishingfilter.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#ifndef LL_PHISHINGFILTER_H
+#define LL_PHISHINGFILTER_H
+
+#include "llsingleton.h"
+#include <string>
+#include <vector>
+
+class LLPhishingFilter : public LLSingleton<LLPhishingFilter>
```
Ansariel, line 34: Why does this have to be a singleton in particular and a class at all? It could just be an encapsulating namespace.

beqjanus:
While I understand the good intention of this PR, I do not think it is at all suitable for FS general use. I have rejected similar proposals in the past (as have LL) because they promote poor security practice by the end-users. OWASP guidance, while not completely ignoring client-side filtering, does not promote it as a line of defence (moreover OWASP imply filtering occurs on input validaton where this is after the effect). 

It has two distinct issues, false positives and false negatives. false positives would result in a user inadvertantly suffering reputational damage and no doubt a spate of ARs that then need to be resolved. False negatives, allow unsafe URLs through under a cloak of apparent safety.

An attacker is able to examine the source code, identify patterns that do not get flagged and suddenly users are being tricked by links that the viewer has not flagged and so "must be safe". No matter how strongly we may word warnings; people will still hold us accountable for not catching those and blame FS for whatever losses they incur. 

I do agree that the right place for such filtering to happen is on the server-side, where LL can not only flag users submitting suspicious URLs for closer review, but block/sanitise the messages. It can so so with reference to adaptive, actively maintained resources and the benefits apply across all users, irrespective of the version of the viewer that they are running. It is unlikely that LL would enter any meaningful discussion with non-LL developers (ourselves included) that mightdivulge any of their security measures, but raising a feature request to promote the conversation and highlight the userbase's concerns would be a decent first step.

Ultimately, there is a simple way to prevent phishing, MFA; beyond that I don't mind things that help expose links, but anything that implies safety through rules or whitelists/blacklists is inherently flawed.

mcgroarty:
I'm afraid I have to agree with the concerns about this training a trust reflex.  Phishing is a real problem, and the current limited server-side filtering isn't enough to protect people. But this will be trivial to bypass, and it will create a false sense of security. Having the rules baked in also means a very slow update cadence. Bad actors will develop workarounds the day of a viewer release, so any protection would be very short-lived.

I like the idea of inline cues - but what about going in the opposite direction and decorating the known-safe links?

Can you think of ways to make official first-party (Second Life, Firestorm) links more prominent? Without creating an eyesore, are there ways to decorate first-party links so that the absence of decoration on them becomes conspicuous?

There could also be an additional, less prominent indicator for known-safe non-SL domains like Discord, YouTube, Primfeed, etc. You could keep that list in a separate GitHub repository so people can request additions/deletions without overwhelming the main project's issue tracker, and so non-devs could be entrusted to help with vetting and list management before QA validates the list format and copies it back to the main project.

minerjr:
Another option would be to use something like [adblock-rust](https://github.com/brave/adblock-rust) library from Brave Browser. Could get the updated lists directly from [easylist][https://github.com/easylist/easylist/tree/master/easylist] GitHub site directly to the users computer and that way Firestorm does not have to maintain the list themselves as well, list it as what library it is using and link to it.

Can use the library to validate the URL and report a with the pop up that it is suspected, or just blocked.

emmelinedevere:
> I like the idea of inline cues - but what about going in the opposite direction and decorating the known-safe links?

> Can you think of ways to make official first-party (Second Life, Firestorm) links more prominent? Without creating an eyesore, are there ways to decorate first-party links so that the absence of decoration on them becomes conspicuous?

I like this line of thinking. Safe domains (secondlife.com, marketplace.secondlife.com, maps.secondlife.com) are trivial to identify, and not subject to the cat-and-mouse game.

Make them blue instead of green?

minerjr:
Need to be careful of only using font color, as some people, including myself are color blind and may need something more like either bold/font style difference, or a different background around the text to make it stand out better. As well, window border changes may be able to help make them stand out better.

Martin-Pitt:
The easiest and safest deescalation is simply dropping the most highly suspicious links (matching very closely to the full official domain name text) from being clickable links.

Anything beyond that like adding other signals like phishing tag or pulling from updated lists I'm not sure about yet in the same line of thinking as beqjanus as being potentially problematic.

Decorating official links with the same colors as used for linden employees could be a good marker. Although it also depends on making sure those being fully safe themselves (no redirect link tricks on any of the domain names or renaming links that stick - marketplace suffers from the latter issue, e.g. `…/p/I-Can-Put-Any-Text-Here-That-I-Want-To/12345678`).

PanteraPolnocy:
This was already discussed in https://jira.firestormviewer.org/browse/FIRE-34469

I, personally, don't believe Firestorm should be responsible for maintaining a whitelist. The responsibility for link safety should ultimately lie with the user rather than the development team. It would be far better to just educate the users: https://www.firestormviewer.org/protecting-your-account-from-phishing-attempts/

That being said...

> Can you think of ways to make official first-party (Second Life, Firestorm) links more prominent? Without creating an eyesore, are there ways to decorate first-party links so that the absence of decoration on them becomes conspicuous?

I think that at one point official links had a small SL's "hand" icon in front of them, added in LLUrlEntrySecondlifeURL / LLUrlEntrySimpleSecondlifeURL.

PanteraPolnocy:
... which actually was not visible if PlainTextChatHistory was set to true (the default) , so: https://github.com/FirestormViewer/phoenix-firestorm/commit/3d2753eb9ead7c0dc262b0813c46e79568b00ee0

PR #225, opened 2026-05-05
Add byte counter to notecard preview

primgineer:
This PR implements live byte counter to the notecard editor UI that displays current byte usage against the maximum allowed size (LLNotecard::MAX_SIZE, 65536 bytes).

### llpreviewnotecard.h
- Added LLTextBox* mByteCounter and std::string mByteCounterTemplate member variables
- Added bool mByteCounterDirty = true member variable
- Declared updateByteCounter() method

### llpreviewnotecard.cpp
- In postBuild(): retrieves the byte_counter widget, caches its template string, and registers a keystroke callback via setKeystrokeCallback that sets mByteCounterDirty = true
- Added updateByteCounter(): gets the current text from the editor, calculates byte size, compares against LLNotecard::MAX_SIZE, sets the label color (white -> yellow at 80% -> red at 100%), and updates the label text with current and max values substituted into the template
- In draw(): checks mByteCounterDirty and only calls updateByteCounter() when set, then clears the flag

### floater_preview_notecard.xml (and locale overrides for ja, ru, zh)
- Added a byte_counter text element with the template string "Bytes: [BYTES] / [MAX]"

## Files Modified
- indra/newview/llpreviewnotecard.cpp
- indra/newview/llpreviewnotecard.h
- indra/newview/skins/default/xui/en/floater_preview_notecard.xml
- indra/newview/skins/default/xui/ja/floater_preview_notecard.xml
- indra/newview/skins/default/xui/ru/floater_preview_notecard.xml
- indra/newview/skins/default/xui/zh/floater_preview_notecard.xml

## Testing
- Compiles without errors and warnings on Windows 10 (VS2022 x64)
- Counts correctly, including 3-byte and 4-byte characters.

## Caveat
- When embedding items inside the notecard the counter doesn't update until a key stroke - setKeystrokeCallback 
  Decided to leave it as it's and not modify the editor class, which is more invasive
  Alternatively, mByteCounterDirty can be dropped and updateByteCounter() can be called every frame from draw(). Performance impact is negligible but perhaps not every likes this approach.

## Use Case
- Provides visual feedback on byte usage, warning the user as they approach the notecard size limit.

### DIFF
--- a/indra/newview/llpreviewnotecard.cpp
+++ b/indra/newview/llpreviewnotecard.cpp
@@ -27,7 +27,7 @@
 #include "llviewerprecompiledheaders.h"
 
 #include "llpreviewnotecard.h"
-#include "llnotecard.h"
+#include "llnotecard.h" // <FS> Byte counter
 
 #include "llinventory.h"
 
@@ -61,7 +61,7 @@
 #include "lllineeditor.h"
 #include "lluictrlfactory.h"
 #include "llviewerassetupload.h"
-#include "lluistring.h"
+#include "lluistring.h" // <FS> Needed for the Byte Counter
 
 // [SL:KB] - Patch: UI-FloaterSearchReplace | Checked: 2010-11-05 (Catznip-2.3.0a) | Added: Catznip-2.3.0a
 #include "llfloatersearchreplace.h"
@@ -79,6 +79,7 @@ LLPreviewNotecard::LLPreviewNotecard(const LLSD& key) //const LLUUID& item_id,
     // <FS:Ansariel> FIRE-24306: Retain cursor position when saving notecards
     ,mCursorPos(0)
     ,mScrollPos(0)
+    ,mByteCounterTemplate("")  // <FS> Initialize empty for Byte Counter
     // <FS:Ansariel> FIRE-29425: User-selectable font and size for notecards
     ,mFontNameChangedCallbackConnection()
     ,mFontSizeChangedCallbackConnection()
@@ -127,14 +128,21 @@ bool LLPreviewNotecard::postBuild()
 
     mEditBtn = getChild<LLButton>("Edit");
     mEditBtn->setCommitCallback(boost::bind(&LLPreviewNotecard::openInExternalEditor, this));
-	
-	// Byte counter
-	mByteCounter = getChild<LLTextBox>("byte_counter");
-	if (mByteCounter)
-	{
-		mByteCounterTemplate = mByteCounter->getText();
-	}
-	mEditor->setKeystrokeCallback([this](LLTextEditor*) { mByteCounterDirty = true; });
+
+    // <FS> Byte counter
+    mByteCounter = getChild<LLTextBox>("byte_counter");
+    if (mByteCounter)
+    {
+        mByteCounterTemplate = mByteCounter->getText();
+        if (mByteCounterTemplate.empty())
+            mByteCounterTemplate = "Bytes: [BYTES] / [MAX]";
+        LLUIString init_text(mByteCounterTemplate);
+        init_text.setArg("[BYTES]", "0");
+        init_text.setArg("[MAX]", std::to_string(LLNotecard::MAX_SIZE));
+        mByteCounter->setText(init_text.getString());
+    }
+    mEditor->setKeystrokeCallback([this](LLTextEditor*) { mByteCounterDirty = true; });
+    // </FS>
 
     // <FS:Ansariel> FIRE-13969: Search button
     getChild<LLButton>("Search")->setClickedCallback(boost::bind(&LLPreviewNotecard::onSearchButtonClicked, this));
@@ -166,28 +174,35 @@ bool LLPreviewNotecard::saveItem()
     return saveIfNeeded(item);
 }
 
-// Byte counter
-// - Neremyn
+// <FS:Neremyn> Byte counter
 void LLPreviewNotecard::updateByteCounter()
 {
     if (!mEditor || !mByteCounter) return;
 
+    mByteCounterDirty = false;
+
+    const size_t MAX_BYTES = LLNotecard::MAX_SIZE;
+
     std::string text = mEditor->getText();
     size_t bytes = text.size(); // Assumes UTF-8 encoding where size() == bytes
 
-    const size_t MAX_BYTES = LLNotecard::MAX_SIZE;
-	
-	auto getColorForByteCount = [MAX_BYTES, bytes]() -> LLColor4
-	{
-		if (bytes >= MAX_BYTES)
-			return LLColor4::red;
-		
-		if (bytes > (MAX_BYTES * 8 / 10))
-			return LLColor4::yellow;
-		
-		return LLColor4::white;
-	};
-	mByteCounter->setColor(getColorForByteCount()); 
+    auto getColorForByteCount = [MAX_BYTES, bytes]() -> LLColor4
+    {
+        if (bytes >= MAX_BYTES)
+            return LLColor4::red;
+
+        if (bytes > (MAX_BYTES * 8 / 10))
+            return LLColor4::yellow;
+
+        return LLColor4::white;
+    };
+    mByteCounter->setColor(getColorForByteCount());
+
+    // Ensure template is valid
+    if (mByteCounterTemplate.empty())
+    {
+        mByteCounterTemplate = "Bytes: [BYTES] / [MAX]";
+    }
 
     LLUIString ui_text(mByteCounterTemplate);
     ui_text.setArg("[BYTES]", std::to_string(bytes));
@@ -195,6 +210,7 @@ void LLPreviewNotecard::updateByteCounter()
 
     mByteCounter->setText(ui_text.getString());
 }
+// </FS>
 
 void LLPreviewNotecard::setEnabled(bool enabled)
 {
@@ -222,11 +238,12 @@ void LLPreviewNotecard::draw()
     bool changed = !mEditor->isPristine();
 
     mSaveBtn->setEnabled(changed && getEnabled());
-	if (mByteCounterDirty) // Byte counter
-	{
-		updateByteCounter();
-		mByteCounterDirty = false;
-	}
+    // <FS> Byte counter
+    if (mByteCounterDirty)
+    {
+        updateByteCounter();
+    }
+    // </FS>
     LLPreview::draw();
 }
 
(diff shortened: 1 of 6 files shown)


indra/newview/llpreviewnotecard.cpp
```diff
@@ -969,4 +1013,4 @@ void LLPreviewNotecard::onFontChanged()
 }
 // </FS:Ansariel>
 
-// EOF
+// EOF
```
Hecklezz: Unnecessary removal of the last line of the file

indra/newview/llpreviewnotecard.h
```diff
@@ -150,4 +158,4 @@ class LLPreviewNotecard : public LLPreview, public LLVOInventoryListener
 };
 
 
-#endif // LL_LLPREVIEWNOTECARD_H
+#endif // LL_LLPREVIEWNOTECARD_H
```
Hecklezz: Unnecessary removal of the last line of the file

indra/newview/llpreviewnotecard.cpp
```diff
@@ -125,6 +127,14 @@ bool LLPreviewNotecard::postBuild()
...
     mEditBtn = getChild<LLButton>("Edit");
     mEditBtn->setCommitCallback(boost::bind(&LLPreviewNotecard::openInExternalEditor, this));
+	
+	// Byte counter
+	mByteCounter = getChild<LLTextBox>("byte_counter");
+	if (mByteCounter)
+	{
+		mByteCounterTemplate = mByteCounter->getText();
+	}
+	mEditor->setKeystrokeCallback([this](LLTextEditor*) { mByteCounterDirty = true; });
```
Hecklezz: 1. Requires Firestorm tagging.
2. You are using tabs instead of spaces for indenting
3. Unnecessary whitespace on line 130 (the new line is fine, but there is a tab there which creates whitespace warnings on linters and git)

indra/newview/llpreviewnotecard.cpp
```diff
@@ -156,6 +166,36 @@ bool LLPreviewNotecard::saveItem()
...
+	};
+	mByteCounter->setColor(getColorForByteCount()); 
+
+    LLUIString ui_text(mByteCounterTemplate);
+    ui_text.setArg("[BYTES]", std::to_string(bytes));
+    ui_text.setArg("[MAX]", std::to_string(MAX_BYTES));
+
+    mByteCounter->setText(ui_text.getString());
+}
+
```
Hecklezz: 1. Requires Firestorm tagging.
2. Mixing use of tabs and spaces for indenting, instead of using only spaces as it should.
3. Unnecessary whitespaces on lines 179, 184, 187, 190 (the new lines are fine, but there is tabs there which creates whitespace warnings on linters and git)

indra/newview/llpreviewnotecard.cpp
```diff
@@ -182,7 +222,11 @@ void LLPreviewNotecard::draw()
     bool changed = !mEditor->isPristine();
 
     mSaveBtn->setEnabled(changed && getEnabled());
-
+	if (mByteCounterDirty) // Byte counter
+	{
+		updateByteCounter();
+		mByteCounterDirty = false;
+	}
```
Hecklezz: 1. Requires Firestorm tagging.
2. You are using tabs instead of spaces for indenting

indra/newview/llpreviewnotecard.h
```diff
@@ -138,6 +141,11 @@ class LLPreviewNotecard : public LLPreview, public LLVOInventoryListener
     LLButton* mDeleteBtn = nullptr;
     LLUICtrl* mLockBtn = nullptr;
 
+	// Byte counter
+	LLTextBox* mByteCounter = nullptr;
+	std::string mByteCounterTemplate;
+	bool mByteCounterDirty = true;
+
```
Hecklezz: 1. Requires Firestorm tagging.
2. You are using tabs instead of spaces for indenting

indra/newview/llpreviewnotecard.cpp
```diff
@@ -60,6 +61,7 @@
 #include "lllineeditor.h"
 #include "lluictrlfactory.h"
 #include "llviewerassetupload.h"
+#include "lluistring.h"
```
Hecklezz: Requires Firestorm tagging.

indra/newview/llpreviewnotecard.cpp
```diff
@@ -27,6 +27,7 @@
 #include "llviewerprecompiledheaders.h"
 
 #include "llpreviewnotecard.h"
+#include "llnotecard.h"
```
Hecklezz: Requires Firestorm tagging

indra/newview/llpreviewnotecard.h
```diff
@@ -99,6 +99,9 @@ class LLPreviewNotecard : public LLPreview, public LLVOInventoryListener
     void updateTitleButtons() override;
     void loadAsset() override;
     bool saveIfNeeded(LLInventoryItem* copyitem = NULL, bool sync = true);
+	
+	// Byte counter
+	void updateByteCounter();
```
Hecklezz: 1. Requires Firestorm tagging.
2. You are using tabs instead of spaces for indenting
3. Unnecessary whitespace on line 102 (the new line is fine, but there is a tab there which creates whitespace warnings on linters and git)

Hecklezz (changes requested):
For the ones I marked requires Firestorm tagging, you can read the FS pull request guidelines and about Firestorm tagging here-
https://github.com/FirestormViewer/phoenix-firestorm/blob/master/CONTRIBUTING.md

Also for the .cpp and .h files, can you please indent with spaces instead of tabs so it is consistent with the vast majority of the rest of the codebase.

Thank you!

Hecklezz (changes requested):
Bug:
mByteCounterDirty is never set to true whenever the asset is fetched from the asset server and loaded later.
Solution: Set it to true in LLPreviewNotecard::onLoadComplete

<details>
<summary>Screenshot showing a notecard after the asset was loaded and it showing the incorrect byte count</summary>

Image

</details>

primgineer:
Fixed the indentations.
Fixed the bug.
Added tags
As for the last lines - I don't know what is causing this. Looked at the original files (cpp and header), both don't have newlines at the end so my editor is adding UNIX-like eol


Notecards with (no copy) permission will show 49 bytes because of the error message - "You do not have permission to view this notecard."

Hecklezz:
GitHub is being really silly right now and keeps giving me "Failed to save comment: An internal error occurred, please try again." whenever I try and review your changes again. 
So while I'm waiting for that to resolve, please have a look over your PR again because there is still a number of cases where tabs are used instead of spaces for indentation, and all of the trailing whitespace causing linter warnings is still in there.
Also, in your latest commit you changed the encoding of llpreviewnotecard.h to UTF-8 BOM for some reason.

You can use git diff --check HEAD~2 on your branch in the terminal and it will show you the trailing whitespaces in red to fix.
As for the tabs instead of spaces, in Visual Studio 2026 (if using another IDE, you would need to look it up), on the .cpp and .h files, at the bottom right of the text editor area, if it says "MIXED" click that and choose Spaces. If it currently says SPC, click it then choose Tabs and then click it again and change it back to Spaces and it should force everything to spaces.
As for the encoding, make sure both say UTF-8 at the bottom right too, and not UTF-8 with BOM like you managed to change one of them to.

Feel free to git commit --amend and force push the changes to save you doing another commit, or if you rather just make another commit, then that works too.

primgineer:
- llpreviewnotecard.h changed back to UTF-8
- Whitespaces removed
- Indentations normalized

Hecklezz:
Alright that looks nicer, and I tested and the bug I had before definitely is fixed now!

One other thing I also noticed, is the #include "llnotecard.h" you added isn't Firestorm tagged (the Firestorm tagging really helps us whenever we merge upstream changes from LL's viewer, that way we know what we have added ourselves and needs to be kept and makes fixing conflicts go faster).

However, I have encountered a new bug-

https://github.com/user-attachments/assets/7a42492e-b534-42fe-a1d7-44a8832d9ae8

The issue appears to be related to this line-
[if (mAssetStatus != PREVIEW_ASSET_LOADED)](https://github.com/FirestormViewer/phoenix-firestorm/pull/225/changes#diff-d78d1cd2f788bc025e45e160d5af9170e04cda14cf72f3595d2a19101adcd558R179)

On some notecards I have, mAssetStatus is set to PREVIEW_ASSET_LOADING even when it has finished loading and is in an editable state, which results in your code setting it to 0/max.
This appears to be a bug in LL's code actually, because they are setting mAssetStatus = PREVIEW_ASSET_LOADING; after it is called onLoadComplete in loadAsset.

Ideally a 1 line change PR (moving the mAssetStatus = PREVIEW_ASSET_LOADING before the gAssetStorage->getInvItemAsset call in loadAsset) should be submitted to LL's viewer, that way every downstream viewer (that includes us) gets it fixed, instead of just us :)

primgineer:
The new changes should address it properly.

- Added the missing tag for #include "llnotecard.h"
- Removed the mAssetStatus guard from updateByteCounter() - the counter now always measures the editor's actual text
- Changed mByteCounterDirty default from true to false - this prevents draw() from calling updateByteCounter() before onLoadComplete. The first real update comes from onLoadComplete (which calls updateByteCounter() directly), and subsequent updates come from keystroke callbacks

The flow is now:
- postBuild() sets initial display to "0 / MAX"
- draw() skips counter (dirty=false) until onLoadComplete or a keystroke
- onLoadComplete calls updateByteCounter() directly - shows real count
- Keystrokes set dirty flag - draw() updates with getText().size()

The previous bug didn't occur after opening and testing all kinds of notecards so hopefully I squashed all the bugs and no there are no regressions.

Hecklezz (approved):
Your changes to work around LL's little oversight appears to work fine on my end with no regressions that I've spotted.

I don't see anything else wrong and all seems to work as intended and is definitely a useful little QoL feature. I approve, but will let another person also review if they spot anything major.

There is 1 tiny little nitpick that is no biggy and no need to change it if you think it's fine, but instead of setting mByteCounterDirty to false every time after calling updateByteCounter, instead you can just set it to false at the top of updateByteCounter.

primgineer:
Updated.

updateByteCounter = false was  removed from onLoadComplete() and draw() and moved at the top of updateByteCounter()
Tested the notecards for the same old bugs just to be sure. Appears to be working as intended.

Thank you for the help!

Hecklezz:
Just letting you know that we plan to merge your feature after our next release, as right now we are focusing on making sure WebRTC and other crucial things are working and fixed for next release. Once that is out this will be merged and will be in the first beta afterwards to be properly tested by a larger userbase.

PR #228, opened 2026-05-08
FIRE-36603: Add missing switch case for enumeration coverage

EricaNebula:
Change introduced in ec92ff9383a37d280875644b74a56e43644e6600 triggers a compiler error on Linux with gcc-15 for missing enumeration coverage in a switch statement in llbutton.cpp

image

The referenced switch:

image

It's missing a case for ::HCENTER (or, if you prefer, a default case). This PR just adds a do-nothing case which satisfies the requirement and allows compilation to continue.

### DIFF
--- a/indra/llui/llbutton.cpp
+++ b/indra/llui/llbutton.cpp
@@ -1238,6 +1238,8 @@ bool LLButton::isLabelTruncated() const
                 text_right -= overlay_width + mImgOverlayLabelSpace;
                 usable_text_width -= overlay_width + mImgOverlayLabelSpace;
                 break;
+            case LLFontGL::HCENTER:
+                break;
         }
     }
 


minerjr:
Hey good catch on that. 

I am going to change the switch statement to a if/else if statements to get rid of the case statement overall. I didn't realize that if  you use a switch statement with an enum, you have to make it complete and not just pick some of the values to work off of. I borrowed the code form the draw method and tried to keep it in tack, but I already had to switch one method call out for the code inside it to make it isLabelTruncated const. So, I should change this from the draw code snip-it to original code style.

Below is what I will put in.
```
        if (mImageOverlayAlignment == LLFontGL::LEFT)
        {
            text_left += overlay_width + mImgOverlayLabelSpace;
            usable_text_width -= overlay_width + mImgOverlayLabelSpace;
        }
        else if (mImageOverlayAlignment == LLFontGL::RIGHT)
        {
            text_right -= overlay_width + mImgOverlayLabelSpace;
            usable_text_width -= overlay_width + mImgOverlayLabelSpace;
        }
```

I'll check it into the code base and can you try it out to verify it works for you. I need to get a Linux VM setup so I can do test compiles of my own. Though I typically used gcc 13 before, so I would have probably still missed it.

In the check in I gave you credit for finding it and your original code suggestion.

Hecklezz:
Closing since an alternative solution was implemented by Minerjr

PR #230, opened 2026-05-12
Fix building RelWithDebInfo due to undeclared identifier isnan

DarlCat:
Simple swap of isnan to llisnan to fix building RelWithDebinfo on Linux. Specifically with Clang 22.1.3

Contributed with permission and blessing from @XenHat

```
indra/llinventory/llsettingsbase.cpp:767:15: error: use of undeclared identifier 'isnan'
  767 |     llassert(!isnan(blendf));
      |               ^~~~~
````


DarlCat:
Added note: this is going to conflict with linden's develop-linux branch anyway, since that has a std::isnan() pass which is currently missing in FS. Felt a code tag may just be more disruptive to that than the conflict itself.

Ansariel:
Why not use std::isnan directly to avoid a merge conflict?

XenHat:
> Why not use std::isnan directly to avoid a merge conflict?

Sure. I just followed the general guidelines and used the ll-provided function because it was the only occurence of isnan and there were plenty of llisnan.

PR #236, opened 2026-06-10
Improved FPS Limiter

XenHat:
This introduces a new hybrid FPS Limiter logic based on sleep/yield and busy loop for maximum stability.
The motivation behind this change is my desire to have a pleasant experience with a VRR (Variable Refresh Rate) monitor, which was difficult with the previous FPS Limiter method:
 - The FPS Limiter was quite inaccurate, and the resulting instability appeared to make every further iteration of the limiting code more unstable by over- and undershooting the target, possibly due to the 1 millisecond accuracy restriction.
 - While undershooting the target wasn't a problem, overshooting was, as it caused stutter from going outside the VRR window.
 - The FPS Limiter was increasingly inaccurate the higher the engine FPS was, likely also due to the millisecond accuracy limit.

The new approach now uses a hybrid busy loop and sleep-based approach to achieve maximum inter-frame stability and consistency.

Observed results seem to reduce the FPS variation from a range between 1 and 200 (!) to ~~1 to 2~~ less than 5 frames.

I have left the old FPS limiter code in place behind a setting to help debugging and testing the code, with the new behavior enabled by default.

I only tested this on Linux x86_64 with AMD Graphics, but it _should_ still work on other platforms.
**Testing required**.

Disclaimer: A local instance of Qwen3 was used to help design this.
I have reviewed the code manually and performed optimizations by hand where it made sense.

Note: Beyond 200 FPS, all things go awry, this limiter is no different. I cannot guarantee the limiter's accuracy in that case.

https://github.com/user-attachments/assets/c59b6e16-5450-4361-95a9-6d36a8d5b243

### DIFF
--- a/indra/newview/llappviewer.cpp
+++ b/indra/newview/llappviewer.cpp
@@ -1984,7 +1984,7 @@ void LLAppViewer::flushLFSIO()
 // <XenHat> More stable FPS Limiter
 #if LL_WINDOWS
 // Waitable timer sleep for precise sub-millisecond waits on Windows.
-static void precise_wait_us(U64 us)
+static void sleep_remaining_us(U64 remaining_us)
 {
     // Use CreateWaitableTimerW for microsecond-precision sleeps.
     HANDLE timer = CreateWaitableTimerW(NULL, TRUE, NULL);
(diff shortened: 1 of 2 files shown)


indra/newview/llappviewer.cpp
```diff
@@ -1993,6 +1978,95 @@ void LLAppViewer::flushLFSIO()
...
+// <XenHat> More stable FPS Limiter
+void inline LLAppViewer::frameLimiterBusyLoop(const U64& sleep_us_remaining)
+{
+    const U64 start_count = get_clock_count();
+    const U64 target_ticks = start_count + (sleep_us_remaining * get_timer_info().mClockFrequency.value() / 1000000);
+    // static LLFastTimer::DeclareTimer FTM_FRAME_LIMITER("Limiter Busy Loop");
+    LL_PROFILE_ZONE_NAMED_CATEGORY_APP("sleep3");
+    while (get_clock_count() < target_ticks)
+    {
+        sched_yield();
```
Ansariel: `sched_yield` is Linux-specific and does not exist on other OS.

XenHat: Well, that's annoying. I'll investigate an alternative for non-Linux OS, perhaps `std::this_thread::yield();` but I'll look at what's available in the viewer already, good catch!
I certainly cannot test other OSes myself.

XenHat: I've submitted an alternative, hopefully this works

Ansariel:
Do these platform-dependent implementations have some actual advantage over using standard C++ functions? You could use `std::this_thread::sleep_for()` if you want to sleep for a defined amount of time.

XenHat:
> Do these platform-dependent implementations have some actual advantage over using standard C++ functions? You could use `std::this_thread::sleep_for()` if you want to sleep for a defined amount of time.

I'm not actually sure. I don't have a windows nor a macOS machine to test the code there, unfortunately. My goal was to have the code as close as possible in functionality to the Linux one. I'll check how well the standard C++ function works :)

XenHat:
I'm not really an expert, but the biggest issue here is the 1ms minimum granularity using the standard sleep function. That's why more fancy methods are needed.

Ansariel:
I am actually wondering if sleeping the main thread is even a good idea. Maybe it would be smarter to skip the render calls but have it do all the other tasks that are part of the main loop.

AFAIR I also heard rumors LL might be cooking up something in this direction. Probably a smart idea to interview Geenz if there is something in the works already.

XenHat:
> AFAIR I also heard rumors LL might be cooking up something in this direction. Probably a smart idea to interview Geenz if there is something in the works already.

I've attempted to contact Geenz before and suggested sharing my implementation in the discord channel and I haven't heard back.
This was PR also meant to give something public to link back to for discussion purpose.

XenHat:
> I am actually wondering if sleeping the main thread is even a good idea. Maybe it would be smarter to skip the render calls but have it do all the other tasks that are part of the main loop.

This would probably be wiser. However, this is currently outside what I remember about the viewer, it's been a while since I touched it.
Suggestions most welcome :)

XenHat:
I'm going to close this since it's unlikely this will be merged.

PR #237, opened 2026-06-15
Native ARM64 (Windows on ARM) viewer port

decaychain:
## Native ARM64 (Windows on ARM) viewer port

This branch builds and runs Firestorm as a **native ARM64 binary on Windows on ARM** (tested on a Qualcomm Snapdragon X / Adreno machine, MSVC ARM64 / VS2022). The viewer starts, logs in, renders the world, and plays sound.

I'm opening this primarily to **get maintainer guidance on how you'd like ARM64 support structured** before I invest in polishing it to merge standard — see "Open questions" at the bottom.

### Base branch
This is based on the **`Firestorm_7.2.4`** release branch (that's the tree I built and tested against), so the diff here is exactly the 5 ARM64 commits and nothing else. If you'd prefer it rebased onto `master`, I'm happy to do that — flagging it because `master` is ~107 commits ahead on a separate line.

### What's in the change

**Source (`indra/`):**
- ARM64 CPU detection (`_M_ARM64` → `LL_ARM64`) and gating of x86-only paths (SSE2 startup check, `_mm_storeu_si128` index packing in `llface`, etc.).
- **GLM SIMD:** `GLM_FORCE_PURE` on MSVC ARM64 — GLM's NEON specializations don't compile under MSVC; scalar math is used.
- **SSE intrinsics:** Microsoft `softintrin` + a small `cmake/arm64_stubs` shim instead of pulling in SSE2NEON on Windows.
- **Boost.Context/Fiber:** `BOOST_USE_WINFIB` (Windows Fiber API), `/guard:signret`, and two ARM64 Pointer-Authentication fixes in Boost itself (heap-allocated activation record; leak the per-thread scheduler at thread-exit to avoid a `SwitchToFiber` fault during teardown). The patched Boost sources are preserved under `arm64-build/boost-patches/`.
- **Audio:** OpenAL Soft backend (`llaudioengine_openal`) reworked off ALUT onto raw ALC + a small WAV loader, since there's no ARM64 FMOD.
- Components without ARM64 binaries are guarded off for this arch: CEF/Dullahan, LibVLC, Discord, GLOD, Growl (with graceful "Not available (ARM64)" version strings).
- libjpeg-turbo built `WITH_JPEG8` (JPEG 80 ABI) to match the viewer's expectations.

**Build tooling (`arm64-build/`, `doc/`):**
- There are no upstream ARM64 Windows autobuild packages, so all 3p libs are built from source for ARM64 and overlaid on the x64 packages (which supply the arch-independent headers). `arm64-build/` preserves the build/staging scripts, the patched Boost sources, and the webrtc m137 customizations; `doc/building_windows_arm64.md` documents the process.

### Testing
- MSVC ARM64, VS2022, Snapdragon X. Configures, builds, launches; login + world render + sound verified.

### Known gaps / honest status
- **Web media / browser (CEF/Dullahan) and LibVLC** are disabled on ARM64 (no ARM64 builds of those upstream).
- **Voice:** webrtc m137 is built for ARM64 and linked, but voice itself is untested.
- I have **not yet** applied Firestorm's `// <FS> ... // </FS>` comment-tag convention around the modified LL code, and there are **no `[FIRE-xxxxx]` JIRA refs** on the commits — happy to add both once we agree on the shape.
- `arm64-build/` scripts have **hard-coded paths** for my build machine; they'd want genericizing (or to live out-of-tree) before merge.

### Open questions for maintainers
1. Preferred base — keep `Firestorm_7.2.4` or rebase onto `master`?
2. Where should the from-source 3p ARM64 build tooling live — in-tree (`arm64-build/`) or a separate repo?
3. Is the `softintrin` + `arm64_stubs` approach acceptable, or would you rather standardize on SSE2NEON (as the Mac-ARM path does)?
4. Any appetite for carrying the Boost ARM64 PAC/winfib patches, or should those go upstream to Boost?

Thanks for taking a look.


beqjanus:
Hi @decaychain 
Thank you for the PR, as you'll see form my replies to the questions, while I very much appreciate the effort here, I don't feel it is well-suited to inclusion in Firestorm directly. (see my closing comment after the questions)

1) Preferred base — keep Firestorm_7.2.4 or rebase onto master?
I would strongly suggest not Firestorm at all but upstream to ll viewer.
2) Where should the from-source 3p ARM64 build tooling live — in-tree (arm64-build/) or a separate repo?
repos are not separated by architecture but are built to a target, as such an arm windows target should be added to each of the 3p repos. This further reinforces my suggestion that this is best pointed to the upstream source at LL, they will be better placed to manage it going forward. 
2) Is the softintrin + arm64_stubs approach acceptable, or would you rather standardize on SSE2NEON (as the Mac-ARM path does)?
Personally, I would suggest standardising, there is no clear value in having a separate implementation here. 
3) Any appetite for carrying the Boost ARM64 PAC/winfib patches, or should those go upstream to Boost?
I have no appetite to carry any of this, least of all any external 3rd party deps. So those would definitiely need to go elsewhere.

I don't feel that firestorm is the right home for this at all. We simply do not have the resources to maintain this going forward, and I am very reluctant to take on such a burden for what remains a very niche platform. I would strongly encourage a discussion with Linden Lab about the best direction to take this, it is possible that they already have plans and if so having a competing/different solution here would be extremely problematic from a maintenance perspective.

PR #243, opened 2026-06-20
Add localized Debug Settings comments support

vedned:
## Motivation

Debug Settings `(Advanced → Debug Settings)` are used to adjust advanced viewer preferences, but their help comments have always been displayed in English, regardless of the language selected in the viewer. This made configuration more difficult for users who are not fluent in English and often required manually translating descriptions to understand what each setting does.

This change introduces support for localized Debug Settings comments, allowing the viewer to display setting descriptions in the user's selected language whenever a translation is available.

## What changes

Previously, all languages used the default English comments defined in app_settings/settings.xml.

Now the viewer loads `skins/default/xui/{locale}/settings_comments.xml` after the session language is initialized and overrides the default comments at runtime.

## Implementation

- Added the `loadLocalizedSettingsComments()` function to `llappviewer.cpp`
- Applies localized comments to both Global and PerAccount settings groups
- Supports loading comment translations from locale-specific `settings_comments.xml` files
- Includes a complete PT-BR translation file in `skins/default/xui/pt/settings_comments.xml` as the first implementation of this system

## Result

With PT-BR as the viewer language, debug settings comments appear in Portuguese using the included `settings_comments.xml`. If no translated file exists for a locale, the previous behavior (English) is preserved.


PanteraPolnocy:
I really appreciate your work on this, both the idea and the execution. However, after sitting on this for a few days, I'm still not entirely convinced it should be merged. Debug settings are intended for advanced folks, not the average viewer user. English language is actually a nice wake-up call after you open debug settings, and you're not a native English speaker.

I know many people just follow the advice of "more experienced" residents and tweak things based on their suggestions - for better or worse. Sometimes our Support team will ask a user to toggle a setting... In those cases, a localized description won't really help them anyway, as the options descriptions are just too technical for them.

That being said, I’m not saying a flat "no" - I'm just still undecided.

vedned:
Thank you for your detailed feedback and for taking the time to consider this carefully.

I fully understand your concern regarding the target audience of Debug Settings and the fact that they are intended for advanced users. I also agree that English can serve as a kind of signal that these options are not meant for casual adjustment.

That said, I would like to clarify the motivation behind the localization proposal. Even within the advanced user group, there are many non-native English speakers who rely on these settings, often based on advice from more experienced residents or support guidance. In these cases, having localized descriptions could reduce misunderstandings and help ensure that changes are made with a clearer understanding of their effects, rather than relying solely on external interpretation or informal translation.

We can adjust the initial proposal to avoid replacing the English defaults or introducing any changes to the current behavior.

Instead, the idea would be to add a button directly within the Debug Settings window, allowing the user to manually select the language they want to view inside that interface.

By default, the interface would remain in English, regardless of the viewer’s language settings. The user would only see another language if they explicitly choose it within that window, with no impact on the rest of the interface or any global system settings.

This approach keeps English as the primary reference and preserves the “warning signal” aspect associated with Debug Settings, while still providing a controlled localization option for users who genuinely need more clarity when interpreting the settings.

I appreciate that you are not rejecting the idea outright, and I am fully open to adjustments or alternative approaches that better fit the intended design philosophy. If there is any possibility to experiment with a small pilot or structured feedback collection, I would be happy to help support that process.

Thank you again for the consideration.

vedned:
To further refine the proposal from an implementation standpoint, this can be treated as a strictly UI-scoped enhancement rather than any form of localization integration into the broader viewer system.

In practical terms, the Debug Settings floater would maintain its existing data source and behavior entirely unchanged, continuing to read and display the default English comments as defined today. The only addition would be an optional runtime override at the presentation layer, applied after the settings data is loaded.

This override would be entirely confined to the Debug Settings window and would not interact with or depend on the viewer’s global language system, skin localization pipeline, or any other UI components outside this context. It would function purely as a per-window display transformation, triggered only when explicitly enabled by the user.
From an architectural perspective, this keeps separation of concerns intact: the settings system remains authoritative and language-agnostic, while the UI layer gains a controlled mechanism for substituting descriptive strings at render time without persistence or side effects.

This also ensures that the existing safety model and “advanced user” intent of Debug Settings is preserved, since the default state remains unchanged and English continues to be the only guaranteed reference without any user action.

vedned:
If useful, I can also provide a short demonstration video of how this would behave in the UI to help clarify the intended behavior before any final decision.

vedned:
## Small fix: guard against empty localized comments

This change mirrors feedback from the parallel upstream work in [secondlife/viewer#5956](https://github.com/secondlife/viewer/pull/5956) (*Localize Debug Settings comments (PT-BR) with English fallback*), which implements the same feature for the official Second Life viewer.

During review of that PR, it was noted that `setComment()` should not be called with an empty string — otherwise a malformed or incomplete entry in `settings_comments.xml` could wipe the original English description. I've applied the same safeguard here in Firestorm.

### What changed

In `loadLocalizedSettingsComments()` (`llappviewer.cpp`), the code now skips entries with an empty comment before calling `setComment()`:

```cpp
const std::string& comment = comment_it->value();
if (comment.empty())
{
    continue;
}
control->setComment(comment);

PanteraPolnocy:
Hm, I see. Well, if this will get pulled upstream, then we cannot really do that ourselves, otherwise merge error will happen next time we try to pull changes from SLV.

vedned:
No problem, I understand. My intention is simply to help the project in the best way possible. If waiting for it to be merged upstream is the better approach, I'm perfectly fine with that. Thank you for the explanation =)

PanteraPolnocy:
I see that it got merged in the Lab's SLV, so we'll just wait for it to come to us here, for maximum compatibility with the upstream code.

PR #251, opened 2026-06-29
Update language Portuguese ( Brazil )

vedned:
Update Portuguese language in preferences


PanteraPolnocy:
Merged, sent a follow up:
https://github.com/FirestormViewer/phoenix-firestorm/commit/2f1205b25f6e08bd6dadadd72b428e89c935fa48

1. Tab and space characters are trimmed by the viewer from the beginning and end of a string, but not in the middle. If you'll let a TAB or SPACE there, it will be displayed - TABs usually as [ ] - empty squares. If you do newlines, please remove all characters from their beginnings.
2. Some values="..." are not needed, and if they change in /en/ - may cause unexpected errors in translations.
3. That one I didn't do, please correct it next time you send a PR: Move notifications to the correct places in the file instead of dumping them all at the end. This will ensure there are no duplicates in the future, due to the fact that during a merge there won't be "ah, a notification seems to be missing here, I'll add it" and person not checking the whole file...

PR #260, opened 2026-07-23
Cap automatic texture memory on Apple Silicon

LisaScheers:
## What changed

- cap automatic texture and vertex-buffer budgeting on ARM64 macOS to one eighth of physical unified memory
- preserve the existing VRAM divisor and automatic-budget behavior on Windows, Linux, and Intel macOS
- preserve explicit manual VRAM budgets
- report the computed automatic budget in viewer diagnostics instead of `Unlimited`
- add focused unit coverage for dedicated and unified-memory calculations

## Why

A profile of Firestorm 7.2.5.81344 on a 48 GB Apple Silicon Mac showed that the viewer detected 38,338 MB as VRAM. With the automatic limiter disabled, the existing controller allowed a target near 30 GB before aggressive texture downscaling. The process reached a 13.9 GB physical-footprint peak, including roughly 8 GB of graphics allocations.

Apple Silicon VRAM is shared system memory, so treating most of the reported value as independently available graphics memory creates avoidable memory pressure. The new cap produces a 6,144 MB automatic budget on the profiled machine before the existing 80% safety target is applied.

The platform-neutral VRAM divisor remains at Firestorm's existing default of `1`; non-unified-memory platforms retain their original budgeting calculation.

## Upstream status

This optimization is also proposed upstream in secondlife/viewer#6034. This Firestorm PR remains a draft pending the upstream review.

## Validation

- compiled the production budget helper and ran all seven regression cases with Nix Clang using `-std=c++17 -Wall -Wextra -Werror`
- covered dedicated VRAM, zero-divisor, minimum-budget, 48 GB unified-memory, low-memory, and missing-memory-data cases
- `xmllint --noout indra/newview/app_settings/settings.xml`
- `git diff --check`

A full viewer build was not run because the profiling host does not have the Linden dependency bundle or an accepted full-Xcode license.

### DIFF
--- a/indra/newview/CMakeLists.txt
+++ b/indra/newview/CMakeLists.txt
@@ -744,6 +744,7 @@ set(viewer_SOURCE_FILES
     lltexturefetch.cpp
     lltextureinfo.cpp
     lltextureinfodetails.cpp
+    lltexturememorybudget.cpp
     lltexturestats.cpp
     lltextureview.cpp
     llthumbnailctrl.cpp
@@ -1599,6 +1600,7 @@ set(viewer_HEADER_FILES
     lltexturefetch.h
     lltextureinfo.h
     lltextureinfodetails.h
+    lltexturememorybudget.h
     lltexturestats.h
     lltextureview.h
     llthumbnailctrl.h
@@ -3012,6 +3014,7 @@ if (LL_TESTS)
     lldateutil.cpp
 #    llmediadataclient.cpp
     lllogininstance.cpp
+    lltexturememorybudget.cpp
 #    llremoteparcelrequest.cpp
     llviewerhelputil.cpp
     llversioninfo.cpp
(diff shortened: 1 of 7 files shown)


Hecklezz:
Okay there is a few problems here:

1. "restore RenderTextureVRAMDivisor to its intended upstream default of 2" is NOT true at all
   This upstream is also defaulted to 1, not 2- https://github.com/secondlife/viewer/blob/7a82e2619598ae6ed1a291d2f7257de8ea55185d/indra/newview/app_settings/settings.xml#L7972
2. Why make a change for Apple Silicon that also impacts Windows and Linux whom don't have the same unified memory? In your changes, you are changing the default of RenderTextureVRAMDivisor to 2 for all OS, despite only Apple Silicon having unified memory (and Intel Macs don't either). 
3. In my opinion, these sorts of viewer optimisations are better suggested to LL in the official Second Life viewer which we fork, this isn't really something that needs to be some Firestorm specific optimisation.
4. Despite all of the above points, IF this were to be merged after more changes to address above concerns, it needs to follow the contributing guidelines and tagging- https://github.com/FirestormViewer/phoenix-firestorm/blob/master/CONTRIBUTING.md

PR #261, opened 2026-07-23
Move texture upload preparation off the main thread

LisaScheers:
## Summary

- precompute fetched-texture alpha classification and pick masks on the existing General worker pool when shared-context texture uploads are disabled
- carry the prepared metadata to the main thread and consume it immediately before the GL upload
- retain the existing synchronous path for explicit formats, unavailable queues, and shared-context uploads
- discard stale preparation when an upload becomes redundant

## Why

A native sample on Apple Silicon placed `LLViewerTextureList::updateImagesCreateTextures()` on roughly 9% of main-thread samples. `LLImageGL::analyzeAlpha()` and pick-mask generation were running as CPU scans inside `setImage()` on the frame-critical path. Firestorm intentionally disables shared-context GL texture uploads for Apple GPUs, but these CPU-only preparation steps do not need the GL context.

## Validation

- `git diff --check`
- deterministic code-path review against the existing alpha-mask and pick-mask algorithms; both paths now share the same alpha classifier
- fallback and stale-result paths verified by inspection

### DIFF
--- a/indra/llrender/llimagegl.cpp
+++ b/indra/llrender/llimagegl.cpp
@@ -42,6 +42,7 @@
 #include "llwindow.h"
 #include "llframetimer.h"
 #include <unordered_set>
+#include <utility>
 
 extern LL_COMMON_API bool on_main_thread();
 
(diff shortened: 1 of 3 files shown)


Hecklezz:

Nitpick: Unnecessary removal of new line at the end of llviewertexture.cpp

PR #262, opened 2026-07-23
Budget incremental reflection probe updates

LisaScheers:
## Summary

- add a token-bucket CPU budget for incremental reflection probe face updates
- measure actual update cost and maintain a smoothed estimate for scheduling
- cap accumulated idle credit so probe work cannot burst without limit
- force an update after 30 deferred frames to guarantee forward progress
- add `RenderReflectionProbeUpdateBudget` with a 1 ms/frame default; `0` restores unthrottled behavior

## Why

A native profile placed `LLReflectionMapManager::update()` / `doProbeUpdate()` on roughly 8% of main-thread samples even with static-only reflections. The existing scheduler performs one incremental face every frame whenever work is pending, without considering the CPU time already consumed by prior faces.

The budget applies to incremental non-realtime updates. Explicit realtime probes retain their current six-face-per-frame behavior.

## Validation

- `git diff --check`
- `xmllint --noout indra/newview/app_settings/settings.xml`
- reviewed credit accrual, measured-cost debit, disabled-budget behavior, and the starvation escape path

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -13034,6 +13034,17 @@ Change of this parameter will affect the layout of buttons in notification toast
     <string>F32</string>
     <key>Value</key>
     <real>2.0</real>
+    </map>
+  <key>RenderReflectionProbeUpdateBudget</key>
+  <map>
+    <key>Comment</key>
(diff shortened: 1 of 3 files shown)


Hecklezz:

Nitpick: Unnecessary removal of new line at the end of llreflectionmapmanager.h

PR #263, opened 2026-07-23
Reduce Apple OpenGL vertex state churn

LisaScheers:
## Summary

- keep the current vertex and index buffers bound across shader transitions on Apple GPUs
- cache which vertex attribute pointer formats are already valid for the bound buffer
- submit only newly required attribute formats, while preserving color/emissive pointer remapping
- retain the existing unbind and setup behavior for non-Apple drivers

## Why

A native Apple Silicon profile showed repeated time in `glVertexAttribPointerARB_Exec`, OpenGL object-name hash lookup/deletion, and dispatch/state update functions. Shader binding currently unbinds both buffers unconditionally, which forces the next draw to bind the same objects and resubmit every active attribute pointer through Apple's OpenGL-on-Metal translation layer.

## Validation

- `git diff --check`
- verified attribute cache invalidation on buffer replacement, buffer changes, and explicit unbind
- verified non-Apple setup behavior remains unchanged
- verified color/emissive combinations preserve the existing pointer source mapping

### DIFF
--- a/indra/llrender/llglslshader.cpp
+++ b/indra/llrender/llglslshader.cpp
@@ -1060,7 +1060,14 @@ void LLGLSLShader::bind()
         {
             sCurBoundShaderPtr->readProfileQuery();
         }
-        LLVertexBuffer::unbind();
+        // Apple's OpenGL-on-Metal layer pays a high price for rebinding buffer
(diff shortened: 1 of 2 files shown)


Hecklezz:

Nitpick: Unnecessary removal of new line at the end of llvertexbuffer.cpp

PR #271, opened 2026-08-18
FIX - FS group title floater does not correctly initialize new region-based buttons

trish-sl:
Minor fix to the group title floater: 
On first init, the buttons for the new region-based roles ("This Region", "Clear Region", etc.) would not correctly initialize after the title list was populated, leaving them incorrectly disabled if you had a title selected. This would force you to click inside the window for their state to refresh proper.


Ansariel:
Looks good to me. Thank you!

PR #272, opened 2026-08-18
Fix LLCallbackMap::map_t incomplete-type build error under GCC/libstdc++

chanayane:
Fixes a GCC/Linux build error in llcallbackmap.h. LLCallbackMap::map_t was declared as a nested typedef using LLCallbackMap as the value type of a std::unordered_map, referencing itself before its own definition was complete. MSVC's STL tolerates this, but libstdc++ requires the value type to be complete when the hashtable node is instantiated, causing a build failure.

Moved the typedef out of the class into a new LLCallbackMapTypes namespace, declared after LLCallbackMap is fully defined. Updated the two call sites in llpanel.h/llpanel.cpp from LLCallbackMap::map_t to LLCallbackMapTypes::map_t (and the associated iterator typedefs) accordingly. No behavior change.

The build error has been introduced by commit 0e687a83b5 — "Optimize various usages of std::map with frequent find access with std::unordered_map" — which swapped LLCallbackMap::map_t from std::map to std::unordered_map, exposing the pre-existing incomplete-type self-reference that only std::map had been tolerating.

### DIFF
--- a/indra/llui/llcallbackmap.h
+++ b/indra/llui/llcallbackmap.h
@@ -39,9 +39,10 @@ public:
     // callback definition.
     typedef std::function<void* (void* data)> callback_t;
 
-    typedef std::unordered_map<std::string, LLCallbackMap> map_t;
-    typedef map_t::iterator map_iter_t;
-    typedef map_t::const_iterator map_const_iter_t;
+    // map_t/map_iter_t/map_const_iter_t are declared after the class (below), not
+    // as nested typedefs: unordered_map<std::string, LLCallbackMap> needs LLCallbackMap
+    // to be a complete type, which it never is within its own body. MSVC's STL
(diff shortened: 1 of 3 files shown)


Hecklezz:
Linux builds perfectly fine though on our Github runners, for myself, Zi and Beq?
Also, modifications to LL's code will need our FS tagging if we merge it anyway.

chanayane:
Must be an issue on my end then. Closing.

chanayane:
As a follow up, the issue was caused by using an ancient GCC version (11.3.0). I had GCC 14.2.0 installed but it wasn't selected.
Adding these solved it.

-DCMAKE_C_COMPILER=/usr/bin/gcc-14 
-DCMAKE_CXX_COMPILER=/usr/bin/g++-14

PR #273, opened 2026-08-19
[FIRE-34340] Issue 1: Fix camtextures and setcam_textures RLVa command for PBR materials.

Atasly:
Fixes [FIRE-34340 First Issue](https://jira.firestormviewer.org/browse/FIRE-34340), i.e.
```
Issue 1:
@camtextures=n does not turn PBR objects grey.
```
This fix also applies to the command `@setcam_textures:[uuid]=<y/n>`

This adds a second pass that iterates all volume prims and overrides the base color UUID on each PBR face's render material. Originals are tracked and restored when the command is removed/cleared.

It matches the Blinn-Phong behaviour.

### DIFF
--- a/indra/newview/llviewerobjectlist.cpp
+++ b/indra/newview/llviewerobjectlist.cpp
@@ -68,7 +68,7 @@
 #include "llmeshrepository.h"
 #include "u64.h"
 #include "llviewertexturelist.h"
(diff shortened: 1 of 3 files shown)


Ansariel:
Looks good at first sight. Let's try it!

PR #275, opened 2026-08-24
[FIRE-34340] Issue 2: Fix camavdist RLVa command to use impostors instead of jelly dolls

Atasly:
Fixes [FIRE-34340 Second Issue](https://jira.firestormviewer.org/browse/FIRE-34340), i.e.
```
Issue 2:
@camavdist:1=n hides avatars or turns them into slightly pixelated tiny blobs
```

This issue has been opened a second one with [FIRE-35778](https://jira.firestormviewer.org/browse/FIRE-35778) which had been resolved with commit [86760e7](https://github.com/FirestormViewer/phoenix-firestorm/commit/86760e7b1faf69b30a01423349d56537901d8cb9)

But I believe the original intent was to have the actual silhouette of an avatar, made pitch black, rendered through an impostor, instead of a jelly doll.

For reference, this is what RLV LSL documentation says:
```
When active, this restriction makes all the avatars beyond <distance> look as if they were visually muted, but colored pitch black. 
```

I could not confirm original behaviour as non PBR FS viewer (6.6.17) already had the issue. I did not go further back in git history to track down the actual intended behaviour.
Any Impostors rendering issues are not fixed by this PR.

Furthermore, I would like to denote that the [Third Issue from FIRE-34340](https://jira.firestormviewer.org/browse/FIRE-34340) has been fixed or is at least not reproductible on my end.

### DIFF
--- a/indra/newview/pipeline.cpp
+++ b/indra/newview/pipeline.cpp
@@ -12027,7 +12027,9 @@ void LLPipeline::generateImpostor(LLVOAvatar* avatar, bool preview_avatar, bool 
 
         LLGLDisable blend(GL_BLEND);
 
-        if (visually_muted || too_complex)
+        // <FS> FIRE-34340-2 RLV silhouettes need a solid color baked into the impostor too
+        if (visually_muted || too_complex || rlv_silhouette)
+        // </FS>
         {
             gGL.setColorMask(true, true);
         }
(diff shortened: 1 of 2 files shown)


indra/newview/pipeline.cpp
```diff
@@ -12049,7 +12052,9 @@ void LLPipeline::generateImpostor(LLVOAvatar* avatar, bool preview_avatar, bool
 
         gDebugProgram.bind();
 
-        if (visually_muted)
+        // <FS> FIRE-34340-2 Use getMutedAVColor() for all muted/complex/silhouette avatars
+        if (visually_muted || too_complex || rlv_silhouette)
```
Ansariel: Why include too complex avatars as well?

Atasly: Per your comment I realized I mistakenly switched two logic blobs while iterating over this issue.
The line you meant is indeed redundant/not needed, but the newly committed one is required for proper RLV silhouettes colorization.

```C++
if (visually_muted || too_complex || rlv_silhouette)
{
    gGL.setColorMask(true, true);
}
```

Apologies for the mishaps.

Ansariel:
Okay, let's try this and see what happens! :)

PR #287, opened 2026-09-14
[FIRE-36105] Add alphabetical and newest-first sort options to outfits menu

TlNKER:
Adds "Sort alphabetically" and "Sort newest first" options to the outfits sort menu in the Appearance floater, alongside the existing "Sort favorites to top" toggle, as requested in FIRE-36105.

Changes:
- Two new LLAccordionCtrl comparators (LLOutfitTabDateComparator, LLOutfitTabFavDateComparator) that sort outfit tabs by creation date, derived from the newest creation date among each outfit folder's direct child items (outfit categories themselves have no reliable creation date in the inventory protocol).
- OutfitListSortOrder is now a bitmask (favorites-to-top and newest-first are independent, combinable toggles) instead of a single mode value.
- Two new menu_item_check entries in menu_outfit_list_sort.xml, wired to the existing Sort.OnSort / Sort.OnEnable handlers.

Tested by compiling and running a staging default-skin build, creating new outfits, and confirming each sort option (and their combination with "favorites to top") reorders the list correctly.


TlNKER:
First contribution and first time doing a PR in my life. If this fails to meet expectation I would tremendously appreciate to know what to do better in the future :-)

Hecklezz:
First of all, your git username which will be public to everyone is currently set to-
"U-DANDELION\\livie" <C:\Users\livie\AppData\Roaming\The Bat!> 
I'd recommend fixing this with the command-
git config user.name "Your username you want here"
Or if you want it global across all repos you wish to push to and not just this repo, use the --global flag-
git config --global user.name "Your username you want here"

Then once you have set your username run this to fix your two commits-
git rebase -i HEAD~2 -x "git commit --amend --reset-author --no-edit"
git push origin fire-36105-outfit-sort --force

Later today if I get a chance, I'll compile the viewer and test the feature and make sure it works right. 

---

There is 1 thing I have concern about though at first glance, which is performance.
If I am understanding the code correctly, every time the list gets sorted, each comparison between two outfits recomputes that outfit's date by walking through all of its items from scratch. Since sorting doesn't just touch each outfit once, this work ends up happening thousands of times per sort instead of once per outfit.

As an example, say a user has 1500 outfits with ~20 items each, one sort does about 31600 folder walks, which is about 630000 individual item checks.
The outfit's date should be cached and only recalculated when that outfit's contents actually change, since the date won't change when sorting (at least it shouldn't...).
Unless I am wrong and misunderstanding what actually would occur in practice, then I think caching would be the right call and would vastly cut down the amount of times the inventory has to be walked.

TlNKER:
Thanks for the feedback Hecklezz! I have updated my username and that will go into the next PR. You are right about the workflow. Can you dismiss this PR? I will try to rework it if I can to touch one link rather than all for each outfits.

Hecklezz:
There is no need to close it, you can continue to use this one since it is the same feature. Just do the rebase and force push I mentioned to update your Username in the commits in this PR, then just do a 3rd commit with the optimisation changes.

TlNKER:
okay, thanks again for the detailed feedback, Hecklezz! Pushed a 3rd commit (f22118d) with a fix.

I originally said I'd try touching just one link per outfit instead of all of them, but backed off that idea once I actually looked at it — getDirectDescendentsOf() doesn't guarantee any particular order for the items it returns, so "the first one" isn't reliably the newest (or oldest). If someone re-saves one old outfit, only that link gets a fresh timestamp; the rest could sit anywhere in the array. Picking just one link risked silently getting the sort wrong some of the time.

So what I went with instead: a cache scoped to a single sort pass. get_outfit_tab_creation_date() now memoizes into an unordered_map that gets primed right before mAccordion->sort() and cleared right after, so each outfit's items only get walked once per sort (the first time that outfit is touched), and every repeat comparison is an O(1) lookup. Since the cache never outlives one sort() call, there's no invalidation logic to get wrong — it just doesn't exist outside of an in-progress sort, so it can't go stale.

For your 1500-outfit/20-item example: thats roughly 1500 folder walks now instead of ~31,600 (so ~30,000 item checks instead of ~630,000) if I got my math right.

Tested on my staging build: sort order is correct for both new options, combining with "favorites to top" still works, and I specifically verified the cache doesn't serve stale data — renaming an outfit (which doesn't touch any item, so its date shouldn't change) leaves it in place, but adding/re-saving an item in an older outfit correctly bumps it to the top on the next sort (swapping to alphabetical and back to newest on top). 

So I suppose the commit is ready for you to go over it :-)

indra/newview/app_settings/settings.xml
```diff
@@ -19019,7 +19019,7 @@ Change of this parameter will affect the layout of buttons in notification toast
     <key>OutfitListSortOrder</key>
     <map>
       <key>Comment</key>
-      <string>How outfit list in Avatar's floater is sorted. 0 - by name 1 - favorites to top</string>
+      <string>How outfit list in Avatar's floater is sorted. Bitmask: 0x1 - favorites to top, 0x2 - newest first instead of alphabetical</string>
```
Hecklezz, line 19033: Instead of modifying the behaviour of LL's existing setting which is subject to be changed by LL at any time, it would probably be best to leave it alone and add our own- "FSOutfitListSortOrder"
Then in the 4 places OutfitListSortOrder is used (which your PR already edits all 4 places), change to using our FSOutfitListSortOrder.
Also don't forget to update the comments that mention OutfitListSortOrder is now a bitmask, as that isn't true anymore.

Then similarly to InventoryAddAttachmentBehavior, just add a- (unused in favour of FSOutfitListSortOrder) to the end of OutfitListSortOrder's description so it is clear to use the new one instead.

indra/newview/skins/default/xui/en/menu_outfit_list_sort.xml
```diff
@@ -21,6 +21,32 @@
...
+   name="sort_newest_first">
+    <on_click
+     function="Sort.OnSort"
+    parameter="newest_first" />
+    <on_check
+     function="Sort.OnEnable"
+     parameter="newest_first" />
+  </menu_item_check>
+
+  <menu_item_separator/>
```
Hecklezz, line 49: Probably best to add FS tagging around this

indra/newview/lloutfitslist.cpp
```diff
@@ -57,6 +57,10 @@
 #include "llviewercontrol.h" // <FS:ND/> for gSavedSettings
 #include "llresmgr.h"
 #include "lltextbox.h"
+
+// <FS:TP> [FIRE-36105] For the per-sort outfit-date memoization cache
+#include <unordered_map>
+// </FS:TP>
```
Hecklezz: Unnecessary and can be deleted. unordered_map is already included in llviewerprecompiledheaders.h which is always included as it is a precompiled header.

indra/newview/lloutfitslist.h
Hecklezz, line 487:
```suggestion

    // <FS:TP> [FIRE-36105] Callback to allow the list to resort itself when outfits change
    void onOutfitItemsChanged(LLWearableItemsList* list, const LLUUID& cat_id);
    // </FS:TP>
```
To fix the bug mentioned (use your own solution if you have a better idea).

indra/newview/lloutfitslist.cpp
Hecklezz:
```suggestion
    // <FS:TP> [FIRE-36105] Use new callback to allow the list to resort itself when outfits change
    //if (!mCategoriesObserver->addCategory(cat_id, boost::bind(&LLWearableItemsList::updateList, list, cat_id)))
    if (!mCategoriesObserver->addCategory(cat_id, boost::bind(&LLOutfitsList::onOutfitItemsChanged, this, list, cat_id)))
    // </FS:TP>
```
To fix the bug mentioned (use your own solution if you have a better idea).

indra/newview/lloutfitslist.cpp
```diff
@@ -864,7 +997,16 @@ void LLOutfitsList::getCurrentCategories(uuid_vec_t& vcur)
...
+    // date-based comparator only walks each outfit's items once (the
+    // first time that outfit is touched during this sort) instead of
+    // on every single pairwise comparison. See sOutfitDateCache and
+    // get_outfit_tab_creation_date() above for details.
+    std::unordered_map<LLUUID, time_t> date_cache;
+    sOutfitDateCache = &date_cache;
     mAccordion->sort();
+    sOutfitDateCache = NULL;
+    // </FS:TP>
 }
```
Hecklezz, line 1021:
```suggestion

// <FS:TP> [FIRE-36105] Callback to allow the list to resort itself when outfits change
void LLOutfitsList::onOutfitItemsChanged(LLWearableItemsList* list, const LLUUID& cat_id)
{
    list->updateList(cat_id);

    S32 sort_order = gSavedSettings.getS32("FSOutfitListSortOrder");
    if (sort_order & OUTFIT_SORT_NEWEST_FIRST)
    {
        sortOutfits();
    }
}
// </FS:TP>

```
To fix the bug mentioned (use your own solution if you have a better idea).

Hecklezz (changes requested):
This approach is definitely a lot better than what it was before and will vastly cut down on inventory walks when people have a crazy amount of outfits. It _could_ go a step further following what I proposed originally where you have a cache that only updates when an outfit change is actually made in your inventory, and that cache will out-live any sort (since your cache only lives for that sort then is nulled). This would require a callback and making sure it is called in all the appropriate places where outfit changes can be made to keep the cache updated. However, your change is a much simpler and less code-impactful and doesn't require any callbacks, and should be far more than performant enough now.

It can always be revisited in the future if users complain about slowdowns compared to before, but I expect it to be perfectly fine as is and suggest to leave it as is for now.

---

I noticed a bug however-
If you save a new outfit with sort newest enabled, they won't be sorted correctly until something else triggers a new sort and then it will be sorted by newest to the top instead of placed at the bottom.
This is a pretty simple to reproduce and fix, and I have added a few code suggestions for one way it can be fixed, but there may be a better way. (If you can think of a better way, then by all means do it your way).

How to reproduce-
1. Sort by newest
2. Create a new outfit named "Test sorting outfit"
3. Observe it get created at the bottom of your outfits list and not sorted to the top as newest, or even sorted by name.

TlNKER:
hi again Hecklezz. thanks for the detailed review! I've pushed a new commit addressing all four points:

The sorting bug: you're right, newly saved outfits weren't sorting into place under "newest first" until something else nudged the list. turns out a brand new outfit has no items in it yet at the moment it shows up, so there's no way to know how "new" it is until its contents actually arrive. I added a callback that fires when an outfits contents change and triggers a fresh sort at that point, so it now lands in the right spot as soon as its items come in instead of waiting around.

I moved off Linden Lab's OutfitListSortOrder and onto our own FSOutfitListSortOrder for the bitmask value. LL's original setting is back to untouched/unused by us, so nothing about our sort options bleeds into it. hope I understood that bit right. 

I cleaned up the FS tagging and updated the comments to reflect the new setting name.

please let me know if anything still looks off :)

indra/newview/lloutfitslist.cpp
```diff
@@ -58,8 +58,9 @@
 #include "llresmgr.h"
 #include "lltextbox.h"
 
-// <FS:TP> [FIRE-36105] For the per-sort outfit-date memoization cache
-#include <unordered_map>
+// <FS:TP> [FIRE-36105] std::unordered_map is used below for the per-sort
+// outfit-date memoization cache; no include needed here, it's already
+// pulled in via llviewerprecompiledheaders.h
 // </FS:TP>
```
Hecklezz: There is really no need for the comments here. The explanation for the use of the unordered_map is explained with the static variable itself. Adding extra comments for an include you aren't even including and why you aren't including it is really overkill.

indra/newview/lloutfitslist.cpp
```diff
@@ -358,7 +363,18 @@ void LLOutfitsList::updateAddedCategory(LLUUID cat_id)
...
+    // known creation date, see get_outfit_tab_creation_date()) at the moment
+    // its tab is added here; this is what actually gives it a date once its
+    // items arrive, so it needs to trigger a fresh sort itself rather than
+    // waiting for some unrelated outfit-list change to do it.
+    if (!mCategoriesObserver->addCategory(cat_id, [this, list, cat_id]()
+    {
+        list->updateList(cat_id);
+        onOutfitItemsChanged(cat_id);
+    }))
+    // </FS:TP>
```
Hecklezz, line 371: Missing LL's original line of code being commented out before your new if (!mCategoriesObserver->addCategory....

indra/newview/skins/default/xui/en/menu_outfit_list_sort.xml
Hecklezz: The tagging around these added lines is still missing :)

Hecklezz (changes requested):
Just a few extra little things but otherwise, everything seems good :)
You can amend and force push your latest commit to address these last things instead of an entire new commit, or you can do a new commit, up to you :)

indra/newview/app_settings/settings.xml
```diff
@@ -19019,7 +19019,18 @@ Change of this parameter will affect the layout of buttons in notification toast
     <key>OutfitListSortOrder</key>
     <map>
       <key>Comment</key>
-      <string>How outfit list in Avatar's floater is sorted. Bitmask: 0x1 - favorites to top, 0x2 - newest first instead of alphabetical</string>
+      <string>How outfit list in Avatar's floater is sorted. 0 - by name 1 - favorites to top</string>
```
Hecklezz: Probably best similarly to the setting "InventoryAddAttachmentBehavior", just add a- (unused in favour of FSOutfitListSortOrder) to the end of OutfitListSortOrder's description so it is clear to use the new one instead for anyone coming who may want to use it in code.

TlNKER:
Updated. Sorry for this whole song-and-dance. still learning!

Hecklezz:
> Updated. Sorry for this whole song-and-dance. still learning!

You are fine haha, I commented on 1 last thing right as you pushed that last commit. I mentioned it in my first review but seems it was missed :)

If you can change that setting description then do-
git add *settings.xml
git commit --amend --no-edit
git push origin fire-36105-outfit-sort --force
And it will add it to your last commit :)

indra/newview/skins/default/xui/en/menu_outfit_list_sort.xml
```diff
@@ -21,6 +21,34 @@
...
+   layout="topleft"
+   name="sort_newest_first">
+    <on_click
+     function="Sort.OnSort"
+    parameter="newest_first" />
+    <on_check
+     function="Sort.OnEnable"
+     parameter="newest_first" />
+  </menu_item_check>
+  <!-- </FS:TP> -->
```
Hecklezz: The closing tags should be below the <menu_item_separator/> just below it, as you added that too

Hecklezz (commented):
And another one, sorry xD

TlNKER:
should be fixed. I am getting cross eyed here xD

indra/newview/lloutfitslist.cpp
```diff
@@ -57,6 +57,7 @@
 #include "llviewercontrol.h" // <FS:ND/> for gSavedSettings
 #include "llresmgr.h"
 #include "lltextbox.h"
+
```
Hecklezz: Extreme nitpick, you have added an extra new line here for no reason 😅

indra/newview/lloutfitslist.cpp
```diff
@@ -57,6 +57,7 @@
...
 #include "lleconomy.h"
```
Hecklezz, line 60: Extreme nitpick, you have added an extra new line here for no reason 😅

Hecklezz (commented):
This is really the last thing I noticed that I'd change, otherwise I'll get Ansariel to do a once over and should be good to merge :)

Hecklezz (approved):

indra/newview/lloutfitslist.h
```diff
@@ -73,6 +73,41 @@ class LLOutfitTabFavComparator : public LLAccordionCtrl::LLTabComparator
...
+ * @class LLOutfitTabDateComparator
+ *
+ * Comparator of outfit tabs sorting newest-created outfit first.
+ */
+class LLOutfitTabDateComparator : public LLAccordionCtrl::LLTabComparator
+{
+    LOG_CLASS(LLOutfitTabDateComparator);
+
+public:
+    LLOutfitTabDateComparator() {};
```
Ansariel: Could just mark the ctor/dtor here and in  `LLOutfitTabFavDateComparator` as `default`.

indra/newview/lloutfitslist.cpp
```diff
@@ -66,6 +66,83 @@ static bool is_tab_header_clicked(LLOutfitAccordionCtrlTab* tab, S32 y);
...
+// can never go stale - it simply doesn't exist outside of an in-progress
+// sort, so there's no invalidation logic to get wrong.
+static std::unordered_map<LLUUID, time_t>* sOutfitDateCache = NULL;
+// </FS:TP>
+
+// <FS:TP> [FIRE-36105] Bits of the "FSOutfitListSortOrder" saved setting.
+// This is a Firestorm-specific setting, not LL's own "OutfitListSortOrder"
+// (which stays a simple 0/1 toggle so LL's own code keeps reading it as
+// before) - see FSOutfitListSortOrder in settings.xml.
+static const S32 OUTFIT_SORT_FAVORITES_TO_TOP = 0x1;
```
Ansariel: Why not `constexpr`?

indra/newview/lloutfitslist.cpp
```diff
@@ -66,6 +66,83 @@ static bool is_tab_header_clicked(LLOutfitAccordionCtrlTab* tab, S32 y);
...
+// Approximate "when was this outfit made" instead using the newest
+// creation date among the outfit's direct contents: saving/updating an
+// outfit creates fresh link items inside its folder, and those links get
+// a real, server-tracked creation date.
+static time_t get_outfit_tab_creation_date(const LLAccordionCtrlTab* tab)
+{
+    LLOutfitAccordionCtrlTab* outfit_tab = (LLOutfitAccordionCtrlTab*)tab;
+    const LLUUID folder_id = outfit_tab->getFolderID();
+
+    // <FS:TP> [FIRE-36105] Serve from the active sort's cache, if any
```
Ansariel: I think the whole function should be tagged and not use the cache part, since it's entirly custom.

indra/newview/lloutfitslist.cpp
```diff
@@ -66,6 +66,83 @@ static bool is_tab_header_clicked(LLOutfitAccordionCtrlTab* tab, S32 y);
...
+    {
+        std::unordered_map<LLUUID, time_t>::const_iterator cached = sOutfitDateCache->find(folder_id);
+        if (cached != sOutfitDateCache->end())
+        {
+            return cached->second;
+        }
+    }
+    // </FS:TP>
+
+    LLInventoryModel::cat_array_t* cats = NULL;
```
Ansariel: Using `nullptr` instead of `NULL` would be nice.

indra/newview/lloutfitslist.cpp
```diff
@@ -66,6 +66,83 @@ static bool is_tab_header_clicked(LLOutfitAccordionCtrlTab* tab, S32 y);
...
+    LLInventoryModel::item_array_t* items = NULL;
+    gInventory.getDirectDescendentsOf(folder_id, cats, items);
+
+    time_t newest = 0;
+    if (items)
+    {
+        for (LLInventoryModel::item_array_t::const_iterator it = items->begin(); it != items->end(); ++it)
```
Ansariel: Why not use range-based iterator?

indra/newview/lloutfitslist.h
TlNKER: LLOutfitTabDateComparator and LLOutfitTabFavDateComparator ctors/dtors now use = default instead of empty bodies.

indra/newview/lloutfitslist.cpp
TlNKER: OUTFIT_SORT_FAVORITES_TO_TOP and OUTFIT_SORT_NEWEST_FIRST are now static constexpr S32.

indra/newview/lloutfitslist.cpp
TlNKER: <FS:TP> tag around get_outfit_tab_creation_date() now wraps the entire function rather than just the cache-check/cache-populate snippets, since the whole function is custom. The cache logic itself is unchanged, assuming that was what you actually meant. If you meant I shouldn't use cache let me know.

indra/newview/lloutfitslist.cpp
TlNKER: Replaced NULL with nullptr throughout the new code in this PR (the sOutfitDateCache declaration/reset and the two locals in get_outfit_tab_creation_date()).

indra/newview/lloutfitslist.cpp
TlNKER: The item-iteration loop in get_outfit_tab_creation_date() is now a range-based for loop.

Ansariel:
Looks ok for me now.

indra/newview/lloutfitslist.cpp
```diff
@@ -66,6 +66,82 @@ static bool is_tab_header_clicked(LLOutfitAccordionCtrlTab* tab, S32 y);
...
+    // Populate the active sort's cache, if any
+    if (sOutfitDateCache)
+    {
+        (*sOutfitDateCache)[folder_id] = newest;
+    }
+
+    return newest;
+}
+// </FS:TP>
+// </FS:TP>
```
Hecklezz: Just this one extra little closing tag that snuck in, in your last commit :)

TlNKER: Forgot to push that! Done now :)

Hecklezz:
Merging this now as it seems all of our performance concerns, bugs and nitpicks are sorted out.
If we get feedback that there is a noticeable performance hit when using the outfits floater, then we can improve it further later doing the suggestion I said earlier where it has a single cache that lives across all sorts that gets updated whenever an outfit update occurs, and not just creating the cache for the life of the sort then nulled again.

PR #288, opened 2026-09-15
[FIRE-34881] Library toggle

TlNKER:
Fixes important part* of FIRE-34881: https://jira.firestormviewer.org/browse/FIRE-34881

Allows user to go into cog wheel > Tabs & Views > Show Library. This toggles whether Library shows in the inventory floater or not. The underpinning objects are still fully accessible to the system, only hidden from the user's view, confirmed that objects from Library still works (Avatar > Avatar Welcome Pack... and all aspects of LSL Bridge works with the Library hidden).

- New debug setting `FSShowLibraryFolder` (default: shown), persisted per-user
- New `LLInventoryPanel::setLibraryFolderVisible()` builds or tears down just   the Library root's folder *view* live, without a restart. The underlying inventory model/data for Library is untouched either way, since other systems (default attachments, outfit lookups, etc.) depend on being able to resolve the Library folder ID regardless of whether its view is shown.
- Newly built folder views need an explicit `arrangeAll()` to actually become visible (`LLFolderViewFolder::addFolder()` has its own `requestArrange()` call commented out), matching the existing pattern used in `openAllFolders()`/`closeAllFolders()`/`setSorter()`.
- New "Show Library" checkbox item added to the gear menu, wired through the existing `Inventory.GearDefault.Custom.Action`/`Check` callbacks.

All changes are tagged `<FS:TP>` per CONTRIBUTING.md, with the original code preserved as comments where replaced as instructed.

(I've decided that was the important part as I was the one raising the ticket to begin with :p)


Hecklezz:
All looks good, however this also hides the Library folder from the texture picker when editing objects, the load sky and water picker in the day/night cycle floater, and the my environments floater under Top Menu->World->Environments->My Environments

Is this intentional of this new toggle, or isn't?

TlNKER:
That is by design. I figured black/white would work best. Either you have it or you don't. (obviously without impacting or effecting viewer performance, so the viewer can still call on the elements in the Library, but it wont show to the user).

Hecklezz (approved):

Ansariel:
AnsaStorm and Vintage don't use this gear menu - it should be added to the menu bar in their main inventory panel XML files, too.

TlNKER:
Good catch! added in the latest push. AnsaStorm and Vintage each have their own panel_main_inventory.xml with the menu items duplicated inline as a visible menu bar rather than referencing the gear menu XML. I added a matching "Show Library" checkbox to both, right next to their existing show_worn_tab/show_favorites_tab entries, wired to the same Inventory.GearDefault.* callbacks and library/toggle_library parameters the default skin's gear menu uses.

Ansariel:
No AnsaStorm Modern, hmmm? 😢

TlNKER:
check again! :P 

AnsaStorm Modern has its own panel_main_inventory.xml too with same inline menu structure as AnsaStorm. Added in the latest push. I also went through every other skin to make sure nothing else was missed. Firestorm and MetaHarper don't override panel_main_inventory.xml at all and StarLight/StarLightCUI do override it but still pull their gear menu from the shared XML rather than duplicating items inline so everything should be covered now. 

(and for some reason I have missed AnsaStorm Modern entirely. that is very pretty!)

Hecklezz:
Going to merge this now as it appears to work fine and now works on all skins haha. 
If we get feedback from our users regarding the question I originally asked about the Library not being accessible in the pickers (like texture picker, etc), then we can make further adjustments to account for that (maybe a checkbox somewhere for if it also hides the Library folder from pickers). But as is, looks good and works.

PR #289, opened 2026-09-18
[FIRE-36987] Add configurable colours for agents loading

TlNKER:
This fixes my ticket on jira: https://jira.firestormviewer.org/browse/FIRE-36987

## Summary
Adds two new preferences so the colour of the avatar loading/bakefail "cloud" particle effect can be customized, instead of it being a fixed color. Exposed as two color pickers under Preferences > Firestorm > Extras.

## Details
- New persisted settings 'FSCloudColorStart' / 'FSCloudColorEnd' store the start and end colors of the cloud particle effect.
- 'LLVOAvatar::applyCloudColor()' applies both to 'sCloud's particle start/end colors, forcing alpha to the stock 0.1 / 0.9 envelope so the effect's shape, burst, scale and timing are unchanged - only the hue is user-configurable.
- Changes to either setting are picked up live via a 'gSavedSettings' signal connection - so no restart required, and takes effect on the next avatar that shows as a cloud.
- Two 'color_swatch' controls (Start / End) added to Preferences > Firestorm > Extras, bound directly to the new settings.

## Testing
- Built and installed to a local staging install with the default skin.
- Verified in Preferences > Firestorm > Extras that both pickers open the color chooser and update the setting.
- Verified an avatar currently rendering as a cloud (loading / bake fail) picks up a color change immediately, with no restart needed.


indra/newview/llvoavatar.cpp
```diff
@@ -1338,7 +1342,46 @@ void LLVOAvatar::initCloud()
...
+{
+    // Same alpha envelope (0.1 -> 0.9) and burst/scale/pattern as the
+    // stock Firestorm cloud.xml - only the start/end hues are user
+    // configurable, so changing colors doesn't change the shape/behavior
+    // of the effect. Alpha is intentionally not exposed to the color
+    // pickers and is forced back to the stock values here.
+    LLColor4 start_color = gSavedSettings.getColor4("FSCloudColorStart");
+    LLColor4 end_color   = gSavedSettings.getColor4("FSCloudColorEnd");
+    start_color.mV[VALPHA] = 0.1f;
+    end_color.mV[VALPHA]   = 0.9f;
```
Hecklezz: Is the alpha values meant to be hardcoded and not changed by the end user? Because FSCloudColorStart and FSCloudColorEnd both store the alpha that the user selects, but then never uses it?

indra/newview/app_settings/settings_firestorm.xml
```diff
@@ -1,6 +1,42 @@
...
+      <key>Value</key>
+      <array>
+        <real>0.945</real>
+        <real>0.522</real>
+        <real>0.714</real>
+        <real>0.9</real>
+      </array>
+    </map>
+    <!-- </FS:TP> -->
+
```
Hecklezz: The default colours have been changed from the original orange colours, to pink.
The default should match what is already in cloud.xml

Hecklezz, line 39: The default colours have been changed from the original orange colours, to pink.
The default should match what is already in cloud.xml

Hecklezz (changes requested):

TlNKER:
Whoops! clear oversight from my end, since I run pink default. Should be fixed now, so both Preferences float shows Red as default selected as well as the actual clouds are the usual red tones defined in cloud.xml until and if the user changes that.

Hecklezz (approved):

indra/newview/llvoavatar.cpp
```diff
@@ -1338,7 +1342,46 @@ void LLVOAvatar::initCloud()
...
+        boost::bind(&onCloudColorChanged, _2));
+    // </FS:TP>
+}
+
+// <FS:TP> [FIRE-36987] Selectable avatar loading/bakefail cloud color
+// Plain-function signal handler - gSavedSettings control signals pass
+// (LLControlVariable* control, const LLSD& new_value, const LLSD& old_value);
+// the new_value itself isn't used since both colors need re-reading either
+// way, so just re-apply from gSavedSettings directly.
+static void onCloudColorChanged(const LLSD& new_value)
```
Ansariel: Why do we need this extra function? `new_value` is never used, so you can just bind directly to `LLVOAvatar::applyCloudColor()` instead.

indra/newview/llvoavatar.cpp
```diff
@@ -1338,7 +1342,46 @@ void LLVOAvatar::initCloud()
...
     // </FS:ND>
-}
+
+    // <FS:TP> [FIRE-36987] Selectable avatar loading/bakefail cloud color
+    // Apply whatever start/end colors are currently saved, on top of
+    // whatever cloud.xml just loaded above, then keep re-applying live if
+    // the user changes either preference later (no restart needed - see
+    // applyCloudColor()).
+    applyCloudColor();
+    gSavedSettings.getControl("FSCloudColorStart")->getSignal()->connect(
```
Ansariel: Should probably store the signal connection and disconnect before if connected, especially since there is a menu option to reload the particle cloud, so you will end up calling `onCloudColorChanged` multiple times.

indra/newview/llvoavatar.cpp
TlNKER: You're right. Removed the onCloudColorChanged() wrapper entirely; the signal now binds directly to LLVOAvatar::applyCloudColor() since new_value was never used.

indra/newview/llvoavatar.cpp
TlNKER: Added sCloudColorStartConnection/sCloudColorEndConnection to store the two signal connections, and initCloud() now disconnects each before reconnecting, so reloading the particle cloud from the menu no longer stacks up duplicate connections.

Ansariel:
I think it's ok now.

Hecklezz:
Going to merge, as there appears to be no other obvious issues and it seems to work fine on my end.

Ansariel:
It only crashed if you didn't use Firestorm mode, so I quickly had to fix it: 2fe01bbec8ef468544372297cb05ef88d1e1f60d

